PR #177, opened 2014-11-20
benchmark incorporation

jonnybazookatone:
It should (in theory) be easy to merge.


jonnybazookatone:
So I do plan to finish this platform in the near future. Given the large changes to bumblebee and how it's grown, I think it is better to keep the benchmarking repository separate. The idea will be you can just easily plug it in next to bumblebee as an extra repo, and run it. Then it shouldn't affect/get in the way of your work flow.

Therefore, I'd suggest you close this ticket and I will open one on my own benchmark repo.

romanchyla:
OK, closing. Can you pls link to that issue from here?

PR #227, opened 2014-12-22
Orcid

romanchyla:
Hi guys,

Please use this branch to make further edits: I've brought it in sync with the latest master. Unfortunately, I didn't have but few hours today, so I didn't do all what I planned. But here is a short list of changes besides merging it with the latest master:

- I have moved most of the new code into modules/orcid (it belongs there naturally, so it's fine)
- moved some functions into more appropriate place (e.g. instead of adding values to the search form and clicking submit, you can accomplish the same by issuing appropriate signal - see orcid_api)
- I've left comments in the code, search for //BC:rca
- in general, it looks good, but the following items _will_ have to be changed:

  - the list_of_things: this class of widgets is general, any code specific to orcid should be inside its own widget (ie. it will not be importing orcid, but orcid widget should extend list_of_things/widget)
 - the results widget should do beehive.getService('OrcidApi') and ask for info about any document it is going to display [I'm realizing this is a very simple situation since Orcid forces you to load all documents at once, so once you have loaded them, we can just keep quering the datastore for every paper we are displaying]
  - tests are not passing (but you know that; additionally some of your changes broke the list-of-things widget tests; I was unfortunately unable to find it) -- needless to say: the tests and coverage will need to pass before it can be integrated

later tomorrow, I can give you a code to display the orcid results in a widget (it is really simple, so you don't need to worry about that part)

btw: I've removed the orcid button from the template (because I wanted to provide an example of how to do it with menus; but I got interrupted - so you won't see it now)


romanchyla:
This PR was closed by #286

PR #270, opened 2015-01-15
Dropdown fix

aholachek:


src/js/page_managers/toc_widget.js
```diff
@@ -67,7 +67,22 @@ define([
     serializeData : function(){
       var data = {};
       data = _.extend(data, this.model.toJSON());
-      data = _.extend(data, {items : this.collection.toJSON()});
+
+      var collectionJSON = this.collection.toJSON();
+
+      viewItems = _.filter(collectionJSON, function(item){
+        if (_.contains(["ShowAbstract", "ShowReferences", "ShowCitations", "ShowCoreads", "ShowTableOfContents"], item.id)){
```
romanchyla: Alex, i'm afraid this won't work - the toc_widget cannot have these hardcoded names in them

what's the intention here? why are they categorized this way?

aholachek: I want to put  the metrics widget in the right spot. I really need to split
the collection into three parts : the "view" part, the "export" part, and
the "analyze" part. Otherwise they will all be under "view". How do I do
that in the current system?

PR #271, opened 2015-01-16
better print css for metrics, but not perfect

aholachek:


romanchyla:
The 'query-template' is missing

but I have a question: why not include the query template into the default template and let the widget always print it? (it already does it anyway, on every render). I thought that you preferred to have the stuff in the templates (rather than being inserted via jquery). I at least tried to do the same when writing new widgets (or updating the old ones)

aholachek:
Ok, sorry about the template, they are now all just in the container template as you suggested.
The issue with jquery was more that stuff in the view should not be toggled in response to user input directly in the html instead of in a model, but I see what you're saying. I think I split them up because I was imagining that you might have to update the different parts of the container template independently, and I didn't want to have to re-render all child views necessarily (in a marionette layout if you re-render the container you destroy the children), but that is never the case here.

PR #279, opened 2015-02-06
three scatter plot options completed, with tests

aholachek:


romanchyla:
Hi Alex,
Please check the following (I'm committing, but you may want to check it after):

 - relative paths ../../../styles vs ../../styles - in container.html
 - can the test data be made smaller? (ideally, just a few cases; will be easier to maintain)
 - can onAll be replaced with view.listenTo('signa')

PR #283, opened 2015-02-12
removed bars, replaced with number vals

aholachek:
Hi Roman, so the numbers need to go back next to the facets, multiple people have asked me to do it. The problem is then the graph bars are both crowded and super redundant. Also, I'm starting to think maybe they weren't a great idea in the first place-- people know how to understand and compare numbers.

Most importantly, Tufte seems to say the same thing here:
http://www.edwardtufte.com/bboard/q-and-a-fetch-msg?msg_id=0001OR .
 Relevant quote:
"People read numbers in tables all the time (see the financial section and sports section of any good newspaper) and they don't need bars to see a bar to understand 1 number."

Can you take a look?

### DIFF
--- a/src/js/page_managers/templates/results-page-layout.html
+++ b/src/js/page_managers/templates/results-page-layout.html
@@ -15,6 +15,7 @@
                 <div class="col-md-2 col-sm-4 hidden-xs s-left-column s-results-column" id="results-left-column">
                     <h2 class="sr-only">Filters for the Current Search</h2>
                     <div class="left-col-container s-left-col-container">
+                        <div class="s-filter-title">Refine by...</div>
                         <div data-widget="AuthorFacet"/>
                         <div data-widget="DatabaseFacet"/>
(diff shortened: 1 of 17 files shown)


romanchyla:
Alex, If we change design every time there is an opposing argument, we won't get anywhere (and for this reason I insisted that there was the 'protection period' - now, when we are out of it, it is crucial to keep steady course)

I don't find that 'multiple people' argument convincing - and I've heard multiple people say that the bard are good, they provide information (but I think they are not useful in case of authors). Facet is not a table, it is a list of names (neither bars, nor numbers are crucial informaiton there - but numbers might be more appropriate in some cases; bars in others - eg. in the database facet the bars are much more telling).

So: yes, I'll take a look. But could I ask to make a stronger case for the numbers? Can we try doing both (and not sacrifice any?) - maybe make numbers go into the background (white font) and appear on hover?

PR #286, opened 2015-02-19
Orcid integration

romanchyla:
This is not yet polished (and I have to remove unused) code, but it is working - and I'm curious about Travis.

discovery.vars.js:
```
orcidClientId: 'APP-P5ANJTQRRTMA6GXZ',
    orcidLoginEndpoint: 'http://sandbox.orcid.org/oauth/authorize',
    orcidApiEndpoint: 'http://localhost:5000'
```

for production, it nees github.com/romanchyla/orcid-service


romanchyla:
OK, so here is the current status and log of changes as they were made:

 - OAuth dance implemented without opening separate windows
 - ORCID api add/set/delete methods correspond to orcid GET/POST/PUT operations (wrote extensive test suite - it was lot of small details to figure out)
 - moved away from XML (ORCID supports JSON for everything)
 - removed almost all the widgets (login, orcid works) as everything can be accomplished by a small extension that is applied on top of existing widgets
 - one main widget (for displaying works from ORCID) remains; bulk operations can be implemented there
 - removed orcid-model-notifier, its usage was unnecessary when the widgets had access to the orcid api anyway
 - implemented (separate) persistent storage mechanism

And to make ORCID integration work, we need a separate webservice (proxy) - http://github.com/adsabs/orcid-service

So, all in all, the original 2K lines of code were condensed into 1K (and the same amount of tests was born)

We'll have to improve the workflow, display of actions, error handling etc - but at this stage, the claims work and we can start playing with it.

PR #292, opened 2015-02-21
more small changes

aholachek:


src/js/page_managers/master.js
```diff
@@ -113,6 +113,14 @@ define([
       this.model = this.view.model;
     },
 
+    ariaDict : {
```
romanchyla: i know that our current system is not ideal, but i think this should be in the template (for now)
instead of hardcoded in the master (cause we may want to use 'master.js' for other apps)

aholachek: So should the template look like:

Now on
{{#if page == "LandingPage" }}
  home page
{{/if}}
{{#if page == "SearchPage" }}
 search results page
{{/if}}
etc...


and if so, do you have any idea how to get this to work since Handlebars
doesn't do string equality? Should we add the helpers library here:
https://github.com/elving/swag ?

romanchyla: I've also found that weird/lacking - so I added it (many commits back), you can do this:

```
{{#compare showDetails "opened"}}
                    <button class="btn btn-default show-details"> <i class="icon-hide-details"></i> Hide details</button>
                    {{else}}
                    <button class="btn btn-default show-details"> <i class="icon-details"></i> Show details</button>
                    {{/compare}}
```

PR #293, opened 2015-02-25
made navbar its own widget

aholachek:
screen shot 2015-03-05 at 8 11 28 am

Hi Roman,
I made the updates we talked about. Above is an updated screenshot.

I still don't have the proper tests. Also, the Orcid api needs to trigger the "orcid ui on" state in the user model (so the user doesn't have to toggle the "orcid mode on" again if they're already signed in), but I'm not sure where that should go


src/discovery.vars.js
```diff
@@ -0,0 +1,61 @@
+define([], function() {
```
romanchyla: pls remove this file from the commit

src/discovery.vars.js.default
```diff
@@ -1,4 +1,4 @@
-define([], function() {
+efine([], function() {
```
romanchyla: careful, this is an error

src/js/components/user.js
```diff
@@ -2,12 +2,62 @@
...
+    },
+
+
+    activate: function (beehive) {
+      this.beehive = beehive;
+      this.pubsub = this.beehive.Services.get('PubSub');
+
+    },
+
+    toggleOrcidUI : function(val){
```
romanchyla: toggle is a name that would well fit the UI element; in the case of the i tis better to call it 'setOrcidValue' -- btw; pls call it more descriptive, something like setOrcidMode

src/js/modules/orcid/extension.js
```diff
@@ -61,28 +63,35 @@ define([
...
       WidgetClass.prototype.processDocs = function() {
         var docs = processDocs.apply(this, arguments);
-        var self = this;
 
-        // add orcid info to the documents
-        var orcidApi = this.beehive.getService('OrcidApi');
-        if (!orcidApi) {
-          return docs;
-        }
+        return this.addOrcidInfo(docs);
```
romanchyla: good change, this is better than my previous version

src/js/page_managers/master.js
```diff
@@ -175,6 +177,13 @@ define([
     handleAriaAnnouncement: function(msg) {
       //template will match the page name with the proper message
       $("#aria-announcement-container").text(AriaAnnouncementTemplate({page : msg}));
+    },
+
+    insertStaticWidgets : function(app){
```
romanchyla: well, static elements are not necessarily limited to navbar only; there should be a better way to select them

src/js/widgets/results/widget.js
```diff
@@ -47,16 +47,26 @@ define([
...
+
+
         if (this.activateResultsExtension)
         {
           this.activateResultsExtension(beehive);
         } // also current this is passed
       },
 
+      onOrcidUIChange: function(val){
+        //how to directly rerender the item views??
```
romanchyla: I think you should go little bit more abstract: the signal is USER_ANNOUNCEMENT

and the action is - re-render everything; the data about the orcid need not be passed - because every item is re-rendered with access to the OrcidApi; the orcid will be displayed just by the virtue of forcing the re-rendering

this functionality is generic; it is not only about orcid

src/js/page_managers/master.js
aholachek: I thought this function could be a catchall for all static widgets
(probably later to include a footer). Could you just add another set of
lines in the function like

 var footer = app.getWidget("FooterWidget");
  $("#footer-container").append(footer.render().el);

Or is that messy?

romanchyla: i though st like the page_managers/controller:assemble function

PR #294, opened 2015-02-26
graphics widget works, the toc widget is now aware of the display_docume...

aholachek:
...nts pubsub call and resets the left nav each time it occurs, and query mediator can handle responses with no content

also, I fixed that annoying bug where sometimes the search bar background is too short and doesn't contain the entire search bar when there is a page transition. (Right now the 'search bar motion' adding/removal is in both discovery_bootstrap.js and master.js, I'm not sure if it should be centralized somewhere else)


src/js/components/query_mediator.js
```diff
@@ -429,9 +429,14 @@ define(['underscore',
       onApiResponse: function(data, textStatus, jqXHR ) {
         var qm = this.qm;
 
+        if (!data && textStatus === "nocontent"){
```
romanchyla: I'll have to test this; i think this won't arrive here (cause is caught by error handler)

Issue #267, opened 2015-01-09
Author facets don't properly filter results


aaccomazzi:
An author search for "Accomazzi, A" returns 174 results:
http://ui.adslabs.org/#search/q=+author%3A%22accomazzi%2C+a%22&sort=date+desc
If I then open the author facet for "Accomazzi, A" I find two entries.  Selecting "Accomazzi, Alberto" and then "limit to" returns the exact same list of papers:
http://ui.adslabs.org/#search/fq=%7B!type%3Daqp+cache%3Dfalse+cost%3D150+v%3D%24fq_author%7D&fq_author=(author%3A%22Accomazzi%2C+Alberto%22)&q=+author%3A%22accomazzi%2C+a%22&sort=date+desc
I assume the problem is simply that the fq query uses the same author parsing and matching rules that aqp uses for q, meaning that instead of doing an exact author search for "Accomazzi, Alberto" it performs the default author query (which is inclusive of all other author variations).

PR #297, opened 2015-03-07
Selected recs + fix for the author filter

romanchyla:
I'm sorry Alex, I mixed up two features into this branch (I didn't realize I haven't pushed the #267) -- but only the last commit is a big one, so hopefully they will not confuse anything.


aholachek:
Hi Roman, Just so I understand correctly, right now there is no way to reset the AppStorage selectedPapers? 

Ultimately, will  the app storage be persistent across sessions? For instance what will happen if a user adds papers, then signs in?

PR #299, opened 2015-03-10
2 orcid issues addressed

aholachek:
I'm not going to lie, I'm not positive I solved the display issue.
But I made the Orcid Mode header darker, so it looks nicer at the cost of conspicuousness.


romanchyla:
here is how i see it
bbb

aholachek:
Ok well I think it looks very artistic that way, no?

No, I will try to figure it out.

aholachek:
Oh wait I think I found it..I will fix it and send it to you after class.

romanchyla:
bingo!

PR #301, opened 2015-03-18
Orcid improvements:

romanchyla:
- added warning before redirecting to orcid
 - modified view on dashboard
 - toggle button now affects display
 - improved orcid_api logic (it will fetch updates when necessary)


romanchyla:
Hi Alex, if you can, please review soon - everything passes, should be easy. thanks

aholachek:
When I tried this locally, the orcid toggle didn't stay on "on" after I authenticated in Orcid--I don't think this is the expected behavior, right?

romanchyla:
thanks, fixed

PR #303, opened 2015-03-19
User-

aholachek:
Hi Roman,

Unfortunately this is still not ready to be merged, but can you take a look at the basic structure and give me some feedback? It is working, but the tests aren't as comprehensive as they should be (although the ones I do have pass), and I still haven't finished dealing with the email confirmation views.


romanchyla:
please rebase with #323

romanchyla:
@aholachek can you pls fix the errors before I start reviewing it? thanks

aholachek:
@romanchyla I'm really sorry, I'm not done yet-- the good thing is there
are only 2 individual tests failing, the bad thing is I still don't have
the second part of the password reset workflow finished and tested. But
it's like 95% done, so it's not ideal but you could still take a look at
it. Sadly I have to leave for the bus right now, otherwise I would fix the
tests.

romanchyla:
OK, i'll review the main ideas without running it

aholachek:
Hi Roman,
Tests are all passing now, (sorry it took so long) although coverage on a few of the tests is just under 80%. Should I keep working until everything is over 80%, or can you look at it now and give me some feedback?

romanchyla:
OK, since the api is giving 500 for /register - I'm going to just look at the code (without testing it). Here is the mix of comments, I'll add more to the sources

- discovery.vars.js: it needs to remain a simple config, should not use Hardened iface, use AppStorage instead
- check in widgets existnce of recaptchaKey and raise approprite errors when not available
- when registration is unsuccessful (it flashes 'error unknown') and then redirects to the same page; that is probably not the intention
- on console: i'm seeing this error:

Uncaught SecurityError: Blocked a frame with origin "https://www.google.com" from accessing a frame with origin "http://localhost:8000".  The frame requesting access has a protocol of "https", the frame being accessed has a protocol of "http". Protocols must match.

This happened after registration submit - but the inter-frame communication makes me nervous. It is going to be troublesome unless done robustly. Comments? Where is it used? Do we *have to* use iframes? (btw: one of the new extensions is putting stuff into global object window - I don't like that at all)

Gruntfile.js
```diff
@@ -422,15 +422,10 @@ module.exports = function(grunt) {
       }
     },
 
-    // concatenates all javascript into one big minified file (of limited 
-    // use for now)
+    // minifying files we need to minify ourselves
     uglify: {
-      options: {
-        banner: "/*! <%= pkg.name %> <%= grunt.template.today('yyyy-mm-dd') %> */\n"
```
romanchyla: pls keep, it will be useful in the future

src/discovery.vars.js.default
```diff
@@ -83,9 +83,13 @@ define([], function() {
       *  orcidApiEndpoint: 'https://api.sandbox.orcid.org/v1.2'
       */
      orcidClientId: 'APP-P5ANJTQRRTMA6GXZ',
-     orcidApiEndpoint: 'https://api.sandbox.orcid.org/v1.2'
+     orcidApiEndpoint: 'https://api.sandbox.orcid.org/v1.2',
+
+
+     // google recaptcha
+     sitekey : "6LfE6AITAAAAALdSy2qxVW4TBqy5RdmREiv-AwlJ"
```
romanchyla: recaptchaKey? - also needs documentation

src/index.html
```diff
@@ -11,12 +11,15 @@
     <link rel="stylesheet" href="./styles/css/smoothness/jquery-ui.min.css" />
     <link rel="stylesheet" href="./styles/css/styles.css" />
 
-<body >
+    <!--how to make this appear only on pages where it's needed?-->
+    <script src="https://www.google.com/recaptcha/api.js?render=explicit" async defer></script>
```
romanchyla: well, yes, try to experiment with loading it dynamically - like bootstrap, as a module dependency

Gruntfile.js
aholachek: hi roman,

thanks for your feedback so far, the frame issue is from the recaptcha. I
know other people online had the same problem, so I'll look again at what
it means.

What should happen on error for register? I show the same page (with a
cleared form)  in case they want to try filling out the form again.

src/js/apps/discovery/router.js
```diff
@@ -13,7 +16,11 @@ define([
     ApiQuery,
     Dependon,
     ErrorTemplate,
-    ApiFeedback
+    ApiFeedback,
+    ApiRequest,
+    ApiTargets,
+    DiscoveryBootstrap
```
romanchyla: this one should not be imported here; it is specific to one mini-app, and it is already available throught the app object tha tthe router receives

if you use some generic method, maybe we should extract it into its own module...

src/js/apps/discovery/router.js
```diff
@@ -83,14 +92,93 @@ define([
...
-        if (view == 'orcid') {
-          this.pubsub.publish(this.pubsub.NAVIGATE, 'orcid-page', subView);
+
+      routeToVerifyPage : function(subView, token){
+
+        var failMessage, failTitle, route, done, fail, request, type;
+
+        if (subView === "register") {
+          failTitle = "Registration failed.";
+          failMessage = "<p>Please try again, or contact <b> adshelp@cfa.harvard.edu for support </b></p>";
```
romanchyla: which reminds me: we should bring the cmment form (as beer has)

src/js/apps/discovery/router.js
```diff
@@ -83,14 +92,93 @@ define([
...
+          route = ApiTargets.VERIFY + "/" + token;
+
+          done = function(reply) {
+            //user has been logged in already by server
+            //request bootstrap
+            this.getApiAccess({reconnect : true});
```
romanchyla: here you need to check the failure/success - it is a promise object

src/js/apps/discovery/router.js
```diff
@@ -83,14 +92,93 @@ define([
...
 
-      userPage: function(view, subView) {
-        if (view == 'orcid') {
-          this.pubsub.publish(this.pubsub.NAVIGATE, 'orcid-page', subView);
+
+      routeToVerifyPage : function(subView, token){
+
+        var failMessage, failTitle, route, done, fail, request, type;
+
+        if (subView === "register") {
```
romanchyla: subView == 'register'

src/js/apps/discovery/router.js
```diff
@@ -83,14 +92,93 @@ define([
...
+            this.getBeeHive().getObject("Session").setChangeToken(token);
+            this.pubsub.publish(this.pubsub.NAVIGATE, 'authentication-page', {subView: "reset-password-2"});
+          };
+
+          failMessage = "Reset password token was invalid.";
+          route = ApiTargets["RESET_PASSWORD"] + "/" + token;
+          type = "GET";
+
+        }
+        fail = function() {
```
romanchyla: this will (eventually) have to be more specific: offering a solution for each of the stages above

src/js/apps/discovery/router.js
```diff
@@ -83,14 +92,93 @@ define([
...
+
+          this.getBeeHive().getService("Api").request(request);
+        },
+
+      orcidPage :function(){
+        this.pubsub.publish(this.pubsub.NAVIGATE, 'orcid-page');
+      },
+
+      authenticationPage: function(subView){
+        //possible subViews: "login", "register", "reset-password"
```
romanchyla: pls list them: if (subView not in ['login', 'register'....]) thrown new Error('Unknown subView')

src/js/apps/discovery/router.js
```diff
@@ -112,6 +200,8 @@ define([
     });
 
     _.extend(Router.prototype, Dependon.BeeHive);
+    _.extend(Router.prototype, DiscoveryBootstrap);
```
romanchyla: nn, discovery bootstrap is only for apps - not for the router

src/js/components/session.js
```diff
@@ -0,0 +1,283 @@
...
+          contentType : "application/json",
+          done : this.loginSuccess,
+          fail : this.loginFail,
+          beforeSend: function(jqXHR, settings) {
+            jqXHR.session = this;
+          }
+        }
+      });
+
+     this.getBeeHive().getService("Api").request(request);
```
romanchyla: return value?

src/js/components/session.js
```diff
@@ -0,0 +1,283 @@
...
+      });
+      this.getBeeHive().getService("Api").request(request);
+    },
+
+    register: function (data) {
+      var base_url;
+      //add base_url to data so email redirects to right url
+      if (!this.test){
+        base_url = location.origin
+        _.extend(data, {verify_url : location.origin + "/#user/account/verify/" + ApiTargets.REGISTER.match(/\/(.*)$/)[1] });
```
romanchyla: better be explicit, not use regex

src/js/components/session.js
```diff
@@ -0,0 +1,283 @@
...
+        query : new ApiQuery({}),
+        options : {
+          type : "POST",
+          data : JSON.stringify(data),
+          contentType : "application/json",
+          done : this.resetPassword1Success,
+          fail : this.resetPassword1Fail
+        }
+    });
+      this.getBeeHive().getService("Api").request(request);
```
romanchyla: all these functions need to return something, even just the promise object

src/js/components/session.js
```diff
@@ -0,0 +1,283 @@
...
+      this.model.set("resetPasswordToken", token);
+    },
+
+    /* private methods */
+
+    loginSuccess: function (response, status, jqXHR) {
+      var promise, pubsub;
+      //reset auth token by contacting Bootstrap
+      //this will log the user in
+      promise = this.getApiAccess({reconnect: true});
```
romanchyla: check the status

src/js/components/session.js
```diff
@@ -0,0 +1,283 @@
...
+      //notify interested widgets
+      pubsub = this.getPubSub();
+      promise.done(function(){
```
romanchyla: and the failure handler?

src/js/components/session.js
```diff
@@ -0,0 +1,283 @@
...
+      pubsub.publish(pubsub.ALERT, new ApiFeedback({code: 0, msg: message, type : "danger", fade: true}));
+      pubsub.publish(pubsub.USER_ANNOUNCEMENT, "register_fail");
+    },
+
+    resetPassword1Success :function(response, status, jqXHR){
+      var pubsub = this.getPubSub();
+      pubsub.publish(this.pubsub.USER_ANNOUNCEMENT, "reset_password_1_success");
+    },
+
+    resetPassword1Fail: function(xhr, status, errorThrown){
```
romanchyla: general remark: these methods are all nicely called - ie. it is clear what they do - you may want to get into habit of documenting them; the expectations under which they get called

src/js/components/session.js
```diff
@@ -0,0 +1,283 @@
...
+      resetPassword1 : "sends an email to account",
+      resetPassword2 : "updates the password",
+      setChangeToken : "the router stores the token to reset password here"
+    }
+
+  });
+
+  _.extend(Session.prototype, Dependon.BeeHive, Dependon.PubSub);
+  _.extend(Session.prototype, Hardened);
+  _.extend(Session.prototype, DiscoveryBootstrap);
```
romanchyla: you should extract the parts of DiscoveryBootstrap into a separate module (or existing mixin)

src/js/components/user.js
```diff
@@ -1,39 +1,90 @@
...
+
+      this.buildAdditionalParameters();
     },
 
     activate: function (beehive) {
       this.setBeeHive(beehive);
-      this.key = this.getBeeHive().getService("PubSub").getPubSubKey();
+      this.setPubSub(beehive);
+      this.key = this.getPubSub().getPubSubKey();
+
```
romanchyla: is there a plan to initialize user settings from theremote storage? (api)

src/js/components/user.js
```diff
@@ -61,23 +112,214 @@ define([
...
-
     //XXX a quick hack
     _persistModel: function() {
-      var storage = this.getBeeHive().getService('PersistentStorage');
+      var beehive = this.getBeeHive();
+      var storage = beehive.getService('PersistentStorage');
       if (storage) {
         storage.set('UserPreferences', this.model.attributes);
       }
+      this.setPubSub(beehive);
```
romanchyla: this should be in activate()

src/js/components/user.js
```diff
@@ -61,23 +112,214 @@ define([
...
+    handleSuccessfulGET : function(response, status, jqXHR){
+      var target = jqXHR.target;
+      this.collection.get(target).set(response);
+      //the change event on the collection should notify widgets
+    },
+
+    handleFailedGET : function(jqXHR, status, errorThrown){
+      // sanity check : if user target is unauthorized, the user isn't actually logged in, so clear the model
+      // not sure if this gets all cases
+      if (jqXHR.target === "USER" && jqXHR.status === 401){
```
romanchyla: ==

src/js/components/user.js
```diff
@@ -61,23 +112,214 @@ define([
...
+      if (!this.callbacks[target]){
+        throw new Error("a post request was made that doesn't have a success callback");
+      }
+      if (this.additionalParameters[target]){
+        _.extend(data, this.additionalParameters[target]);
+      }
+      this.composeRequest(target, "POST", data);
+    },
+
+    /*return read-only copy of user model(s) for widgets: accessible through facade(2 of 3)*/
```
romanchyla: nice

src/js/components/user.js
```diff
@@ -61,23 +112,214 @@ define([
...
+    hardenedInterface: {
+      completeLogIn : "sync user object with database",
+      completeLogOut: "clear user object",
+      isLoggedIn: "whether the user is logged in",
+      postData: "POST new values to user endpoint (params: endpoint, data)",
+      getUserData: "get a copy of user data currently in the model for an endpoint, or all user data (params: optional endpoint)",
+      getUserName: "get the user's email before the @",
+      isOrcidModeOn : "figure out if user has Orcid mode activated",
+      setOrcidMode : "set orcid ui on or off",
+      setUser : "set the username to log the user in and fetch his/her info"
```
romanchyla: it is ok, but normally i prefer the following modus operandi:

widgets have access to read-only methods
IFF they have access to write methods, the write method will check whether the widget is allowed to call it
OR use pubsub.publish signal and let mediator (can be the user object itself) listen to the request and update data

src/js/mixins/form_view_functions.js
```diff
@@ -0,0 +1,117 @@
...
+
+  //some functions to be used by form views which auto-validate
+  var formFunctions = {
+    //for the view
+    //call in onRender method to activate the recaptcha
+    //otherwise it won't work
+    activateRecaptcha: function (sitekey) {
+      var that = this;
+      //for testing, ignore
+      if (window.grecaptcha) {
```
romanchyla: can we convince that google library to not use globalobjects?

src/js/page_managers/master.js
```diff
@@ -91,7 +94,7 @@ define([
           model.attributes.numAttach += 1;
 
           //scroll up automatically
-          window.scrollTo(0,0);
+//          window.scrollTo(0,0);
```
romanchyla: nice! are we sure it is no longer needed?

src/js/services/api.js
```diff
@@ -73,21 +75,37 @@ define([
...
       if (query && !(query instanceof ApiQuery)) {
         throw Error("Api.query must be instance of ApiQuery");
       }
 
+      //can pass in query parameter (instance of ApiQuery) or just pass data directly in the data parameter
+      //if you pass in an api query, it overrides data, otherwise just pass in data for the ajax request as expected
+      if (query){
+        data = options.contentType === "application/json" ? JSON.stringify(query.toJSON()) : query.url();
+      }
+      else if (data){
```
romanchyla: isn't data always undefined?

src/js/services/api.js
```diff
@@ -99,12 +117,16 @@ define([
         type: 'GET',
         url: u,
         dataType: 'json',
-        data: options.contentType === "application/json" ? JSON.stringify(query.toJSON()) : query.url() || "{}",
+        data: data,
         contentType: 'application/x-www-form-urlencoded',
-        headers: {"X-BB-Api-Client-Version": this.clientVersion},
+        headers: {"X-BB-Api-Client-Version": this.clientVersion, 'X-CSRFToken' : this.csrf},
```
romanchyla: the mediator, that i mentioned, should handle this - csrf tokens are not needed otherwise and don't need to be in the default api

src/js/widgets/alerts/widget.js
```diff
@@ -50,9 +53,20 @@ define([
...
 
       events: {
-        'click #alertBox button.close': 'close'
+        'click #alertBox button.close': 'close',
+        'click button[data-dismiss=modal]' : 'closeModal'
+
       },
 
+      closeModal : function(){
+        debugger;
```
romanchyla: nuke

src/js/widgets/navbar/widget.js
```diff
@@ -150,7 +180,15 @@ define([
       }
     },
 
-    handleUserAnnouncement: function (msg, data) {
+    //to set the correct initial values for signed in statuses
+    setInitialVals : function(){
+      var user = this.beehive.getObject("User");
+      var orcidApi = this.beehive.getService("OrcidApi");
+      this.model.set({orcidModeOn : user.isOrcidModeOn(), orcidLoggedIn:  orcidApi.hasAccess()});
```
romanchyla: right now, the ui is green, even if i have no acccess to orcid api

Gruntfile.js
romanchyla: maybe a more specific error message, displayed to the user

romanchyla:
Alex, it is a big PR, please expect more comments once I can run it - e.g. I just glossed over the tests - and there will be even more feedback once we put it into hourly. But it all looks nice and clean to me, some changes here and there will be needed, but nothing major. Well done!

Gruntfile.js
aholachek: Ah ok. I changed the uglify to create a new copy of backbone.min, since new backbone doesn't come with backbone.min-- should I just change the uglify back to what it was, or keep both versions?

romanchyla: How is it related? Uglify minimizes everything it can find,  BTW I have
slap fixed the backbone in other commits,  that you already integrated
during rebase

romanchyla: Egh, slap is my phone being incomprehensible :)

src/js/apps/discovery/router.js
aholachek: where is it available? since it is a mixin I assumed it wasn't available with getBeeHive().getObject/getService or something similar. Or should the two functions getApiAccess and on Bootstrap be their own  service called "apiConnect" or something?

Also, in a related question, when would you need to call getApiAccess and *not* use the reconnect: true option?

src/js/components/session.js
aholachek: does it need a return value? It is making a post that will either succeed or fail and then be handled by the correct callbacks.

src/js/components/session.js
aholachek: so then it would break if the name of the "register" target changed, right?

romanchyla: doesn't the regex just match the register without front '/' ? that is the same as '/#user/account/verify' + REGISTER

aholachek: Right now the entries in the api_targets file all have
"accounts/[endpoint]", and I just wanted the name of the endpoint

romanchyla: imo one reason more to not use the regex - since if you change the 'accounts/foo' to be '/accounts/foo' explicit is better than implicit (assumption about the shape of the register target)

aholachek: ok, I understand now

romanchyla:
you forgot to add the js/mixins/api_access.js, thanks

aholachek:
Oh no, sorry, I added it.

romanchyla:
Alex, these updates did brake tests; see travis - pls also rebase against master

PR #322, opened 2015-04-01
Added ability to query ADS to crosscheck orcid ids

romanchyla:
Alex, pls wait with this one (if it passes, that's it) - I'd like to add tests.


romanchyla:
Hi Alex, it is now ready for review/integration. Thanks

aholachek:
Hi Roman,
I noticed you added the callback adding a handlebars operator in the discovery config under the "callback" parameter. Is that function the place to add any modifications to the libraries we need to make? For instance I need to register something for the backbone-validation library, and this change is used by multiple widgets.

romanchyla:
Yes, if it is app-global, then that is the good place.

PR #328, opened 2015-04-09
Network additions

aholachek:
The main changes included 
- adding a graph to show the times that different groups were active, measured by the amount of papers they produced
- adding an ability to "hop" from author to author in the author network, with a back button that allows the user to go to the previous network
- standardizing author and paper network to use bootstrap layout classes so they are responsive, and just improving the styles in general

### DIFF
--- a/test/mocha/js/widgets/network_widget.spec.js
+++ b/test/mocha/js/widgets/network_widget.spec.js
@@ -1689,16 +1689,25 @@ define([
     var networkWidget = new NetworkWidget({networkType: "author", endpoint: "author-network", helpText: "test"});
     $("#test").append(networkWidget.view.el);
 
+    var apiQuery = new ApiQuery({q: "star"})
+
     //this should show not enough data template
-    networkWidget.processResponse(new JsonResponse(testDataSmall));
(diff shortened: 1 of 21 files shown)


romanchyla:
the summary view seems to be very very tiny (i've regenerated less,  made no difference)

screenshot - 04102015 - 04 01 02 pm

and the question marks disappeared (instead it points to a help page) - i think that is a regression, since the user will leave the page. could we have the question marks (and link to the help pages only from the short help?) i belive this will give you also more space to use

btw: this PR would benefit from descriping the nature of the changes, it is not obvious from the code

test/mocha/js/widgets/network_widget.spec.js
```diff
@@ -1758,6 +1773,9 @@ define([
     networkWidget.view.graphView.model.set("selectedEntity", accomazziNode);
     $("#test").find("button.filter-add").click();
 
+
+    debugger;
```
romanchyla: tbremoved

romanchyla:
Alex, can you pls rebase this one? thanks

aholachek:
Done!

PR #330, opened 2015-04-10
removed question mark

aholachek:
I'm not sure how this mistake was made... anyway, have we decided to definitively not do pushstate for the foreseeable future?

### DIFF
--- a/src/js/widgets/abstract/templates/abstract_template.html
+++ b/src/js/widgets/abstract/templates/abstract_template.html
@@ -15,9 +15,9 @@
         <ul class="list-inline">
             {{#each authorAff}}
                 {{#if @last}}
-                <li class="author"><a href="#search/?q=author%3A{{this.[2]}}&sort=date+desc">{{this.[0]}}</a><span class="affiliation hide"> (<i>{{this.[1]}}</i>)</span></li>
+                <li class="author"><a href="#search/q=author%3A{{this.[2]}}&sort=date+desc">{{this.[0]}}</a><span class="affiliation hide"> (<i>{{this.[1]}}</i>)</span></li>
                 {{else}}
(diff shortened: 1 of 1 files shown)


romanchyla:
we can do pushstate if we spend (if we configure nginx to handle certain routes) - i'm not too keen on it

but, the templates *have to* use hashes, because that is the smallest common denominator (not all browsers can handle pushstate, but all can do hashes)

aholachek:
I don't think templates have to use hashes.

I thought the whole point of the code we used to have that caught all
internal links and sent them to the router was that we could use standard
urls in a non-pushstate enabled browser.

I copy pasted it here: (it's been removed from bumblebee for now since I
thought we weren't using pushstate)

// All navigation that is relative should be passed through the navigate
  // method, to be processed by the router. If the link has a `data-bypass`
  // attribute, bypass the delegation completely.
  $(document).on("click", "a[href]:not([data-bypass])", function(evt) {
    // Get the absolute anchor href.
    var href = { prop: $(this).prop("href"), attr: $(this).attr("href") };
    // Get the absolute root.
    var root = location.protocol + "//" + location.host + app.root;

    // Ensure the root is part of the anchor href, meaning it's relative.
    if (href.prop.slice(0, root.length) === root) {
      // Stop the default event to ensure the link will not cause a page
      // refresh.
      evt.preventDefault();

      // `Backbone.history.navigate` is sufficient for all Routers and will
      // trigger the correct events. The Router's internal `navigate` method
      // calls this anyways.  The fragment is sliced from the root.
      Backbone.history.navigate(href.attr, true);
    }
  });

romanchyla:
if hashes are there, it works always. if hashes are not there (in the templates), it works only if you run it with properly configured nginx/webserver; and we don't have that

try to load (fresh) a page /search/?query=foo - as opposed to #search/?query=foo and observe what happens with relative links, css styles etc

and secondly, it is much simpler (the rules are straightforward) to have the hashes there rather than not have them (cause if there are not, you will have to know/expect/verify two places - router and the webserver)

PR #336, opened 2015-04-13
finished recommender

aholachek:
Hi Roman,

The recommender widget is simple but it brings up two issues that affect other widgets:

1. There doesn't need to be an api query in an api request, for some endpoints (recommender, graphics) the "target" is enough because it has the bibcode in it
2. The 404 throwing an error issue that I tried to address with the graphics widget pull request. But that pull request checks if the 404 comes from the graphics endpoint, so it wouldn't take into account the 404 issue from the recommender endpoint or citation helper endpoint.

I removed the requirement for the query in the discovery_mediator, default_request, and api.js. I commented out the test in api_request.spec.js that checks the query validation, should I just remove it?

### DIFF
--- a/src/js/apps/discovery/router.js
+++ b/src/js/apps/discovery/router.js
@@ -9,6 +9,7 @@ define([
     'js/components/api_targets',
     'js/mixins/api_access',
     'js/components/api_query_updater'
+
   ],
   function (
     $,
@@ -86,7 +87,6 @@ define([
 
       view: function (bibcode, subPage) {
 
-<<<<<<< HEAD
         // check we are using the canonical bibcode and redirect to it if necessary
         var q, req, self;
         self = this;
@@ -113,21 +113,11 @@ define([
         }});
 
         this.pubsub.publish(this.pubsub.EXECUTE_REQUEST, req);
-=======
-          if (!subPage) {
-            return this.pubsub.publish(this.pubsub.NAVIGATE, 'abstract-page', "#abs/"+bibcode+"/abstract");
-          }
-          else {
-            var navigateString = "Show"+ subPage[0].toUpperCase() + subPage.slice(1);
-            return this.pubsub.publish(this.pubsub.NAVIGATE, navigateString, "#abs/"+bibcode+"/"+subPage);
-          }
-        }
-        this.pubsub.publish(this.pubsub.NAVIGATE, 'abstract-page');
+
       },
->>>>>>> added routes for abstract page, changed discovery mediators method of showing loading views
 
 
-  routeToVerifyPage : function(subView, token){
+      routeToVerifyPage : function(subView, token){
 
         var failMessage, failTitle, route, done, request, type,
           that = this;
@@ -142,11 +132,11 @@ define([
             //request bootstrap
             this.getApiAccess({reconnect : true}).done(function(){
               //redirect to index page
-              that.pubsub.publish(this.pubsub.NAVIGATE, 'index-page');
+              that.pubsub.publish(that.pubsub.NAVIGATE, 'index-page');
               //call alerts widget
               var title = "Welcome to ADS";
               var msg = "<p>You have been successfully registered with the username</p> <p><b>"+ reply.email +"</b></p>";
-              that.pubsub.publish(this.pubsub.ALERT, new ApiFeedback({code: 0, title : title, msg: msg, modal : true, type : "success"}));
+              that.pubsub.publish(that.pubsub.ALERT, new ApiFeedback({code: 0, title : title, msg: msg, modal : true, type : "success"}));
             }).fail(function(){
               //fail function defined below
               fail();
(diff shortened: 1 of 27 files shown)


src/js/components/default_request.js
```diff
@@ -31,7 +31,7 @@ define([
       return true;
     };
     var allowedAttrs = {
-      query: function(v) {return v instanceof ApiQuery},
+      query: basicCheck,
```
romanchyla: hmm, but query should always be an instance of api-query

src/js/services/api.js
```diff
@@ -83,11 +83,23 @@ define([
...
+      var q;
       var query = request.get('query');
       if (query && !(query instanceof ApiQuery)) {
         throw Error("Api.query must be instance of ApiQuery");
       }
 
+      if (query && options.contentType == "application/json" ){
+        q = JSON.stringify(query.toJSON());
+      }
+      else if (query){
```
romanchyla: this will be break if I set query to {}

src/js/widgets/recommender/widget.js
```diff
@@ -40,49 +39,41 @@ define([
...
-          self.trigger('page-manager-event', 'widget-ready', {'isActive': true, numFound: data});
+        this.loadBibcodeData(bibcode).done(function() {
+          self.trigger('page-manager-event', 'widget-ready', {'isActive': true});
         });
       }
     },
 
     loadBibcodeData : function(bibcode){
 
+      this.deferredObject =  $.Deferred();
```
romanchyla: use of the deferred escapes me here; especially in the second condition, it will not be resolved - this seems incorrect

test/mocha/js/components/api_request.spec.js
```diff
@@ -24,16 +24,17 @@ define(['js/components/api_request', 'js/components/multi_params', 'js/component
       expect(r.set('target', 'bar')).to.be.OK;
       expect(r.get('target')).to.eql('bar');
 
-      // it accepts only api query objects
```
romanchyla: even if you made it optional, you should keep/modify this test - because one *can* set a query

src/js/components/default_request.js
aholachek: but what if there isn't a query at all?

aholachek: should it be
query: function(v) { if (v) {return v instanceof ApiQuery}} ?

src/js/services/api.js
aholachek: I wanted to add the option to *not* have a query, and just use the target-- however if there is a query, it should be an instance of ApiQuery

src/js/components/default_request.js
romanchyla: v !== undefined

romanchyla:
@aholachek let me know when ready for integration of this one

aholachek:
Hi @romanchyla, I'm trying to add loading views (the spinning wheels) to widgets on the abstract page--I tried to add the proper handler in the discovery_mediator, but it doesn't seem to be working. Can you take a look?

romanchyla:
i'm still confused here - the loadBibcode() doesn't need a promise; it can just send an EXECUTE_REQUEST

other source of confusion (to me) is the use of special signal to turn spinning wheel on (i know you wanted me to look into it, but I think we'll need a different logic - so I haven't tried); I think the spinning wheel may  be needed for any widget which calls EXECUTE_REQUEST, in that case the solution would be to update `query_mediator.executeRequest()' as follows:

```
ps.publish(self.pubSubKey, ps.FEEDBACK, new ApiFeedback({
            code: ApiFeedback.CODES.SEARCH_CYCLE_STARTED,
            cycle: [oneId]
          }));
this._executeRequest(apiRequest, senderKey)
 .done(
ps.publish(self.pubSubKey, ps.FEEDBACK, new ApiFeedback({
            code: ApiFeedback.CODES.SEARCH_CYCLE_FINISHED,
            cycle: [oneId]
          }));
)
.fail(
ps.publish(self.pubSubKey, ps.FEEDBACK, new ApiFeedback({
            code: ApiFeedback.CODES.SEARCH_CYCLE_FAILED,
            cycle: [oneId]
          }));
)
```

src/js/wraps/discovery_mediator.js
```diff
@@ -53,6 +53,24 @@ define([
       }
     };
 
+    handlers[ApiFeedback.CODES.BIBCODE_DATA_REQUESTED] = function(feedback){
```
romanchyla: as we discussed, this hardcoded list of widgets won't fly far

better to make spin every widget that makes EXECUTE_QUERY request (see my previous comment)

src/js/widgets/recommender/widget.js
```diff
@@ -40,54 +39,43 @@ define([
...
-        return this.deferredObject.promise();
+        this.trigger('page-manager-event', 'widget-ready', {'isActive': true});
       }
       else {
         this._bibcode = bibcode;
-        this.deferredObject =  $.Deferred();
-
-        var target = "services/recommender/" + bibcode;
-
+        var target = ApiTargets.RECOMMENDER + "/" + bibcode;
```
romanchyla: i;m not sure about it, but some bibcodes may contain weird characters - and if they do, this will fail the request

src/js/apps/discovery/router.js
```diff
@@ -78,21 +76,21 @@ define([
...
-        if (bibcode) {
+        if (bibcode){
           this.pubsub.publish(this.pubsub.DISPLAY_DOCUMENTS, new ApiQuery({'q': 'bibcode:' + bibcode}));
-        }
 
-        if (!subPage) {
-            navigateString = 'abstract-page';
-        }
+          //for loading views
+          this.pubsub.publish(this.pubsub.FEEDBACK, new ApiFeedback({code : ApiFeedback.CODES.BIBCODE_DATA_REQUESTED}));
```
romanchyla: if the spinning wheel starts turning for every widget which did request data, this won't be necessary

alternatively, nice idea would be to slightly modify the logic - to send a list of widgets, that should be marked as 'loading' - however, there is a problem with 'unmarking' them

src/js/wraps/discovery_mediator.js
aholachek: ok, i think i was confused-- you want to totally change the way all widgets show spinning wheels, not just for the abstract page-- so I will have to change the logic for search results widgets too.

PR #339, opened 2015-04-14
changed styles to work with firefox

aholachek:
fixed firefox issue


aholachek:
this is styling mostly so I will just merge it

PR #365, opened 2015-05-04
limit is 1000 now

aholachek:


romanchyla:
is metrics service working fine with 1000 recs?

aholachek:
it worked ok when I made about 5 different requests for 1000 records -- I dont know if that means it will work all the time though

PR #432, opened 2015-06-25
reorganized front page to be more in line with bumblebee aesthetic, a…

aholachek:
…dded footer, added hover info on search bar


romanchyla:
please see #434 - i have mixed/confused the two, but they contain overlapping features

romanchyla:
@aholachek please rebase, this was overlpaing with #434 - and now I can't merge. I would have lots of troubles with deciding which of the css changes should be incorporated

PR #434, opened 2015-06-26
ui enhancements

aholachek:


romanchyla:
I'm imagining there is a lot of suggestions that arrive at your table and that you try to accommodate - and it is the unhappy reality of the UI that each of us think we know what is best. 

So here is my take: I'm hesitating here, because I almost feel the people behind your shoulder pointing with their finger at what could be 'improved'; good intentions, but they focus on details and don't realize the effect on the whole

Please zoom out and look at the pictures from far away: 

 - is it intentional to make the page 'heavier'? The text font is bolder and the colour is darker (I did run `grunt less`)
 - the numbers in facets are definitely intentional, but has it improved the situation? (in my opinion, we are again sinning in the direction of "trying to show everything".... cause it is annoying to have to hover over the name to get the counts...; but seriously - do we care so much about this detail as to break the flow of the page? And if those numbers are crucial to see, then perhaps we should remove the bars completely?)

- the histogram (on the right) saved some space to loose it in other areas (and alignment is broken)

current:
image

new changes:
image


And compare these two: the new page looks much better (and the teaching link is great), but I'd like to point out the prominent space that "NASA + ADS + Bumblebee" occupies. It grew by 200%. Why do we mention Bumblebee at the top at all? I don't know the real reasons, but I would like to say we should be invisible; not interfere with the purpose for which people are there. I believe these little details are very important. It is not about us (developers), it is about them (scientists, users)

current:
image

updates:
image

PR #437, opened 2015-07-01
added fix for IE

aholachek:


src/js/components/session.js
```diff
@@ -127,13 +127,13 @@ define([
       var base_url;
       //add base_url to data so email redirects to right url
       if (!this.test){
-        base_url = location.origin
+
         _.extend(data, {verify_url : location.origin + "/#user/account/verify/register" });
       }
       else {
-        base_url = "location.origin"
+        _.extend(data, {verify_url : "location.origin" + "/#user/account/verify/register" });
```
romanchyla: concatenating strings? not location.origin + ..... ?

romanchyla:
- this logic fails when bbb is installed in /foo
 - maybe we'll have to use modernizr (we already use es5-shim in tests)
 - it shows we are long overdue with automated testing against different browsers
 
Can you confirm the two string concatenantion is correct? We can then commit, but mark it temporary.

aholachek:
I use the string only for the tests, when location.origin is unknown. Is
there a better way to do it?

I don't think modernizr handles javascript (it handles only html5 and css3
I think?)

Anyway for location.origin I used what was suggested here on the modernizr
site:
https://github.com/Modernizr/Modernizr/wiki/HTML5-Cross-Browser-Polyfills

romanchyla:
Let me rephrase it: you want to concatenate two literal values and that is
how it should work.

If yes, then we can merge it (and mark temp)

aholachek:
ok, i think it is good. when i was looking at it again i realized i should have made it more concise by using the ternary operator, though, so please wait until the latest version passes travis first.

PR #442, opened 2015-07-06
Feature/issue#34/openurl-generator-iau

jonnybazookatone:
This should address points 4 and 5 in adsabs/issues/issues/34, the ability to create OpenURLs from the meta-data of the document.

The following changes and additions were applied:
  1. An object (or mixin?) was created `OpenURLGenerator`, which sits in `/js/mixins/openurl_generator.js`, which contains all the logic for parsing the meta data and creating a "ContextObject" to be used for openURL generation.
  2. A test suite for this mixin was included and is located in `/js/mixins/openurl_generator.spec.js`, and "should" cover most cases, if not all hopefully... The test suite was included in the `/js/core-suite.js` test suite.
  3. The LinkGeneratorMixin was updated to import the OpenURLGenerator, and the logic included in the link generation, replacing the old code that simply went to ADS classic. The logic is essentially: if its an electronic paper, the user has a library link, and there is no ADS scan, then generate an openURL and place it in the links list.
  4. Tests were added to the LinkGeneratorMixin test suite.

Apologies for the mass of commits, I am sure they can be rebased, but I avoided that currently as I didn't want to mess up my repo entirely :-).

Of course comments are welcome and encouraged if something is really terrible.


test/mocha/js/mixins/link_generator_mixin.spec.js
```diff
@@ -151,9 +150,195 @@ define(['js/mixins/link_generator_mixin'],
...
+          '{"title":"", "type":"preprint", "instances":"", "access":"open"}',
+          '{"title":"", "type":"electr", "instances":"", "access":"open"}'
+        ];
 
+        // Check that an openURL is NOT created
+        var output = mixin.getTextAndDataLinks(stub_links_data, stub_meta_data.bibcode, stub_meta_data);
+        expect(_.where(output.text, {title : "Publisher Article"})[0]["link"]).to.not.contain("url_ver");
+      });
+
+      it("should not generate an openURL if there is an ADS scan available", function() {
```
romanchyla: what if you want to generate links to both the scan and the resolver? (isn't that actually st we want to do?)

src/js/mixins/openurl_generator.js
```diff
@@ -0,0 +1,225 @@
...
+         * Generates a class that can be used to create an OpenURL and a ContextObject for OpenURLs
+         * OpenURL semantics have been followed using the standard Z39.88-2004, v0.1 and v1.0 should
+         * both be supported, in principle.
+         *
+         * @param {object} metadata - should contain the docs from a Solr response
+         * @param {string} link_server - library link server to be used as the base of the OpenURL
+         */
+
+        // Fixed values
+        this.baseURL = link_server;
```
romanchyla: i'm reading (almost) sequentially :)

if this is a generator, it might be advisable to collect these fields into a separate object, because e.g. 'baseUrl' might be already there in the class that you extend by the generator (or you could protect these variables by extracting them into the scope outside of the OpenURLGenerator class)

test/mocha/js/mixins/link_generator_mixin.spec.js
jonnybazookatone: I followed the logic from [ADS2.0](https://github.com/adsabs/adsabs/blob/083d70ba399e965eda8c7ab70489b0192bd77a43/adsabs/modules/abs/templates/macros/abstract.html#L126) as per Alberto's suggestion. If this wasn't what was wanted, then yeah we can include both the scan and the resolver.

romanchyla: @aaccomazzi - do we want to give users both links? to the scanned version, as well as the openurl link? Or is it XOR situation?

aaccomazzi: It's XOR: the purpose of the openurl link is to allow a user to read something which is behind a paywall but for which the user may have access via an institutional subscription.  We don't need to do this for records which are open access (which the ADS scans are).

PR #445, opened 2015-07-07
added preferences widget, and myads to user component

aholachek:


src/js/widgets/search_bar/search_bar_widget.js
```diff
@@ -459,8 +459,8 @@ define([
             }
           });
 
-          if (operator && operator === "references(" ){
-            query.set("sort", "first_author asc")
+          if (operator && operator === "references(" || operator === "instructive(" ){
```
romanchyla: operator == "instructive"

src/js/widgets/resources/widget.js
```diff
@@ -77,11 +88,26 @@ define([
 
     },
 
+    handleUserAnnouncement : function(event, target, model_data){
```
romanchyla: I'm not sure this is necessary; since you have exposed the User object, the logic can be simplified to

var user = beehive.getObject('User');
if (user.hasOpenUrl()) {
  // generate the open url link
}

romanchyla:
Alex, please check the coverage (and unless under stress, we should not lower down the limits in that Gruntfile.js - rather the opposite ;))

src/js/widgets/search_bar/search_bar_widget.js
adsabs-beer-user: This seems wrong.  The sorts should be:

instructive() - relevance desc
reviews() - relevance desc
trending() - relevance desc
references() - first_author asc
citations() - pubdate desc

AFAICT all of these were correct in the code except for reviews()

-- AA


On Tue, Jul 7, 2015 at 6:21 PM, Roman Chyla <notifications@github.com>
wrote:

> In src/js/widgets/search_bar/search_bar_widget.js
> <https://github.com/adsabs/bumblebee/pull/445#discussion_r34098185>:
>
> > @@ -459,8 +459,8 @@ define([
> >              }
> >            });
> >
> > -          if (operator && operator === "references(" ){
> > -            query.set("sort", "first_author asc")
> > +          if (operator && operator === "references(" || operator === "instructive(" ){
>
> operator == "instructive"
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/adsabs/bumblebee/pull/445/files#r34098185>.
>



-- 
Dr. Alberto Accomazzi
Program Manager
NASA Astrophysics Data System - http://ads.harvard.edu
Harvard-Smithsonian Center for Astrophysics - http://www.cfa.harvard.edu
60 Garden St, MS 83, Cambridge, MA 02138, USA

src/js/widgets/resources/widget.js
aholachek: so every time it renders it should get the openurl? Right now it listens for a change and only requests it 1x. I can change it if it's better to get it anew every time.

romanchyla: yes please, it is in memory already anyway, it will be simpler

romanchyla: btw: once this change is done, the PR can be merged, no need to review again

PR #473, opened 2015-07-12
router uses replace=true

aholachek:
Hi Roman,

This fixes the back button problem.
I'm not totally sure if the change in src/js/components/query_mediator.js is right for all situations, can you take a look and tell me what you think? 

Basically you need to use {replace: true} for all cases when the call to navigate is coming from the router, so you don't put 2 entries in history when there should be only 1.

### DIFF
--- a/src/js/components/navigator.js
+++ b/src/js/components/navigator.js
@@ -60,6 +60,7 @@ define(['underscore',
        * Responds to PubSubEvents.NAVIGATE signal
        */
       navigate: function(ev, arg1, arg2) {
+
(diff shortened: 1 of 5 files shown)


src/js/components/navigator.js
```diff
@@ -84,11 +85,14 @@ define(['underscore',
           this.handleTransitionError(transition, e, arguments);
         }
 
+        //router can communicate directly with navigator to replace url
```
romanchyla: note for later: the PR was merged, but this seems like a hack

PR #478, opened 2015-07-13
temporary fix for token problem

aholachek:
Hi Roman,

I must have caused this error when I refactored  the user settings to use the TOC widget.
I will work on streamlining the user object and user settings widget today/tomorrow.

Can you check to see that you can get a token, and everything works ok?


romanchyla:
tested, it works

PR #487, opened 2015-07-15
refactored user object, user settings widget, and increased coverage

aholachek:


src/index.html
```diff
@@ -10,14 +10,10 @@
...
     <link rel="stylesheet" href="./styles/css/styles.css" />
-    <link href='http://fonts.googleapis.com/css?family=Montserrat:700' rel='stylesheet' type='text/css'>
-
-
     <!--favicon stuff-->
-
     <link rel="icon" type="image/png" sizes="32x32" href="./styles/img/favicon.ico">
 
-
+</head>
```
romanchyla: wow, i never noticed! :)

src/js/apps/discovery/navigator.js
```diff
@@ -231,11 +231,10 @@ define([
...
 
           if (loggedIn){
-            //redirect to preferences
-            app.getObject('MasterPageManager').show("SettingsPage",
-              ['UserSettings']);
-            app.getWidget("UserSettings").setSubView("preferences");
-            this.route = "#user/settings/preferences"
+            //redirect to index
+            app.getObject('MasterPageManager').show('LandingPage');
+            var q = app.getObject('AppStorage').getCurrentQuery();
```
romanchyla: what if there is none?

src/js/components/user.js
```diff
@@ -367,45 +249,147 @@ define([
...
       setOrcidMode : "set orcid ui on or off",
-      setUser : "set the username to log the user in and fetch his/her info",
-
       getOpenURLConfig : "get list of openurl endpoints",
-      setMyADSData : "",
+      getUserData : "myads data",
+      setUserData : "POST user data to myads endpoint",
+      generateToken : "PUT to token endpoint to make a new token",
+      getToken : "GET from token endpoint",
+      deleteAccount : "POST to delete account endpoint",
```
romanchyla: i also like this pattern (of calling api through the object), however we need to put security checks there (at least)

we can't allow every widget to call deleteAccount(); thoughts?

src/js/widgets/navbar/widget.js
```diff
@@ -237,12 +237,14 @@ define([
...
 
-      var user = this.getBeeHive().getObject("User");
-      var orcidApi = this.getBeeHive().getService("OrcidApi");
+      var orcidApi = this.getBeeHive().getService("OrcidApi"),
+          user = this.getBeeHive().getObject("User");
 
-      if (msg === "user_info_change" && arg2 === "USER") {
-        //if user logs out, username will be undefined
-        this.model.set("currentUser", this.getBeeHive().getObject("User").getUserName());
+      if (msg == "user_signed_in") {
```
romanchyla: let's make those things CONSTANTS; cause a reviewer is getting confused whether you wanted to send a literal value or it is just a mistake; with additional advantage you can change the signal values at will, ie.

```
if (msg == User.USER_SIGNED_OUT) {
....
```

src/js/widgets/user_settings/widget.js
```diff
@@ -292,170 +298,100 @@ define([
...
    },
 
    activate : function(beehive) {
      this.beehive = beehive;
      this.pubsub = beehive.Services.get('PubSub');
-     _.bindAll(this, ["handleUserAnnouncement", "handleOutsideNavigate"]);
+     _.bindAll(this, ["handleUserAnnouncement"]);
      this.pubsub.subscribe(this.pubsub.USER_ANNOUNCEMENT, this.handleUserAnnouncement);
-     this.pubsub.subscribe(this.pubsub.NAVIGATE, this.handleOutsideNavigate);
+     this.view.getToken = this.beehive.getObject("User").getToken;
```
romanchyla: pls change to a funciton call
```
function() {
  return this.beehive.getObject("User").getToken();
}
```

the reason is to avoid loitering

test/mocha/js/components/user.spec.js
```diff
@@ -131,115 +135,122 @@ define([
...
-
-      User.prototype.broadcastChange.restore();
+       s.restore();
+       s2.restore();
 
      });
 
 
-   it("allows widgets to post data to user endpoints", function(){
+   it("allows widgets to post/put data to user endpoints (user data, change email, change password, delete account, get token)", function(){
```
romanchyla: when we allow widgets post data, we have few problems to solve:

- synchronization (widgetA, widgetB; both update the same value; which gets firts; how do they know about each other?)
- error checking (widgetA can put in something that confounds widgetB and makes it fail)

user stuff is similar to queries; potentially many widgets are interested in knowing/subscribing to changes

i guess we need to look into this problem closer

test/mocha/js/components/user.spec.js
```diff
@@ -274,94 +285,34 @@ define([
...
-     sinon.stub(User.prototype, "broadcastChange");
-
-     sinon.spy(u, "completeLogOut")
-
-     var publishStub = sinon.stub(u.pubsub, "publish");
-
-     u.redirectIfNecessary = sinon.stub();
-
-     //get success
-     u.handleSuccessfulGET({api_token : "foo"}, "success", {target: "TOKEN"});
```
romanchyla: sorry, does this mean that the api is returning 'success' when the operation was successful?

that seems weird; the success should be coded int the HTTP response (200 OK)

romanchyla: am i missing something? is there a reason to do it that way?

src/js/components/user.js
aholachek: only the user_settings_widget should ever need to call delete account, whats the best way to ensure that happens?

src/js/widgets/navbar/widget.js
aholachek: Right now User publishes USER_ANNOUNCEMENT to let widgets know of multiple events, and the possibilities are:
"user_signed_in"
"user_signed_out"
"user_info_change"
"OrcidUIChange"
(and also right now there is "logout_success" but that can probably be replaced with user_signed_out)

I agree it's a lot, so you are saying I should have an object that contains constants and the descriptions of what the events do? Does that by itself make things less confusing, or am I missing something else?

test/mocha/js/components/user.spec.js
aholachek: When the $.ajax call fails, it passes these parameters : jqXHR.fail(function( jqXHR, textStatus, errorThrown ) {});
The jqXHR.textStatus contains the actual message from the API. 

According to the jquery docs, the text status parameter is " a string categorizing the status of the request ("success", "notmodified", "nocontent", "error", "timeout", "abort", or "parsererror"). "

Does this answer the question or is there something I'm missing?

romanchyla: post mortem note

i find it confusing to pass 'success' to the handler which only handles success (it is not clear where this value come from - by just looking at the funciton)

src/js/widgets/navbar/widget.js
romanchyla: the constants yes; cause it tells me this value is something that the code recognized (and it is not something that the api sent) - this on itself removes the ambiguity

romanchyla:
i'm going to merge this with comments and TODO's

PR #491, opened 2015-07-16
added recaptcha, changed endpoint to slack endpoint

aholachek:
Roman, before you merge this, I have a test question (a question about tests, not a question testing you.)

Ok. I was under the impression that calling done() in setTimeout meant that you could go over 2000 ms. But when I do that in navbar.spec.js, I get a timeout. Am I doing something wrong or do I misunderstand what done() does? (It turns out I don't need such a long setTimeout in this particular case, but for future reference).

### DIFF
--- a/src/js/widgets/navbar/widget.js
+++ b/src/js/widgets/navbar/widget.js
@@ -5,6 +5,8 @@ define([
   'hbs!./template/feedback',
   'js/components/api_query_updater',
   'js/components/api_query',
+  'js/components/api_request',
+  'js/components/api_targets',
   'bootstrap'
 
 ], function(
@@ -14,6 +16,8 @@ define([
   FeedbackTemplate,
   ApiQueryUpdater,
   ApiQuery,
+  ApiRequest,
+  ApiTargets,
   Bootstrap
   ){
 
@@ -109,11 +113,13 @@ define([
 
     onRender : function(){
 
+      var that = this;
+
       if (!this.formAttached){
         //attach modal
         $("body").append(FeedbackTemplate());
 
-        var $modal = $("#feedback-modal"), that = this;
+        var $modal = $("#feedback-modal");
 
         function clearForm () {
           $modal.find(".modal-body").html($(FeedbackTemplate()).find("form"))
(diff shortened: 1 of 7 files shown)


aholachek:
this is failing coverage for recaptcha manager now, I will add a test

romanchyla:
done() is a signal (a callback) to mocha that an asynchronous test can be closed; the global/local mocha timeouts still apply independently; so yes, if you wait too long, mocha will fail you (or fail on you?...)

src/js/widgets/navbar/widget.js
```diff
@@ -127,7 +127,7 @@ define([
           var $form = $(e.target);
           e.preventDefault();
           $.ajax({
```
romanchyla: this won't work in production, can we move it to the api_targets?

test/mocha/js/components/recaptcha_manager.spec.js
```diff
@@ -25,6 +29,41 @@ define([
...
+
+      var minsub = new (MinSub.extend({
+        request: function(apiRequest) {
+        }
+
+      }))({verbose: false});
+
+      var fakeAppStorage = {getConfigCopy : function(){return {recaptchaKey : "here_is_a_fake_key"}}};
+
+      minsub.beehive.addObject("AppStorage", fakeAppStorage);
```
romanchyla: just a comment: in the not-so-near future, we'll have to come up with a system of loading objects (some will depend on some others; but i don't wnat to complicate things unnecessarily); maybe it will be enough to ask one/few mediators (?)

aholachek: do you mean "not so distant"?

romanchyla: yes, yah...(stutters)..that o-one!

romanchyla:
do we need this new feature in production? i'm tempted to answer: no - and merge it to master only

adsabs-beer-user:
but we want feedback!

adsabs-beer-user:
from people at the IAU!

romanchyla:
but the feedback works, you receive it to the email, don't you?

and also, is the feedback/slack available in production API?

On Mon, Jul 20, 2015 at 11:25 AM, adsabs-beer-user <notifications@github.com
> wrote:

> from people at the IAU!
>
> On Mon, Jul 20, 2015 at 11:25 AM, Holachek, Alexandra <

romanchyla:
had a look at the feedback endpoint, it allows for feedback to be sent from anywhere - which may not be ready for production (think spam), so i'm waiting for https://github.com/adsabs/adsws/issues/73

PR #500, opened 2015-07-20
bbb is now somewhat mobile/tablet friendly

aholachek:
Hi Roman,
I put a delegated event handler for showing/hiding a hidden nav in main.js. Is there a better place to put it?


src/js/apps/discovery/main.js
```diff
@@ -58,6 +58,15 @@ define(['config', 'module', 'analytics'], function(config, module, analytics) {
           app.start(Router);
           pubsub.publish(pubsub.getPubSubKey(), pubsub.APP_STARTED);
 
+          $("body").on("click", "button.toggle-menu", function(e){
```
romanchyla: yeah, we should find a better place for this

how about the index.html?

aholachek: so no jquery allowed?

romanchyla:
jquery allowed, but it should be the same that we use normally (which is hard to synchronize)

how about page_managers/tempaltes/master-page-manger.html ?

that should work and have jquery available by default

PR #503, opened 2015-07-21
f

aholachek:
Hi Roman,

This is adding the abstract widget to the abstract page.

It took a while because I was confused about tests -- the minsub's Api object is not accessible to widgets because it doesn't have a getHardenedInstance function, so I added an option to have that (in case the widget both uses pubsub and uses Api.request directly).

This is what it should look like, with non-relevant information removed:
screen shot 2015-07-21 at 12 26 45 pm

### DIFF
--- a/src/js/bugutils/minimal_pubsub.js
+++ b/src/js/bugutils/minimal_pubsub.js
@@ -69,8 +69,8 @@ define(['underscore', 'backbone',
       this.pubsub.debug = true;
       this.beehive.addService('PubSub', this.pubsub);
       var self = this;
-      this.beehive.addService('Api', options.Api || {
-        request: function(req, context) {
+      var Api = {
+        request: function (req, context) {
           self.requestCounter += 1;
           if (!context) {
             context = req.get('options');
@@ -80,12 +80,15 @@ define(['underscore', 'backbone',
             console.log('[MinSub]', 'request', self.requestCounter, response);
           }
           var defer = $.Deferred();
-          defer.done(function() {
-              context.done.call(context.context, response);
+          defer.done(function () {
+            context.done.call(context.context, response);
           });
           defer.resolve(response);
           return defer.promise();
-        }});
+        }
+      };
+      this.beehive.addService("Api", options.Api || Api);
+
       this.beehive.addObject('QueryMediator', new QueryMediator({
         recoveryDelayInMs: 0,
         shortDelayInMs: 0,
--- a/src/js/widgets/metrics/widget.js
+++ b/src/js/widgets/metrics/widget.js
@@ -7,6 +7,7 @@ define([
   'js/components/json_response',
   'js/components/api_request',
   'js/components/api_query',
+  'js/mixins/dependon',
   'js/mixins/user_change_rows',
   'hbs!./templates/metrics_metadata',
   'hbs!./templates/metrics_container',
@@ -28,6 +29,7 @@ define([
   JsonResponse,
   ApiRequest,
   ApiQuery,
+  Dependon,
   UserChangeMixin,
   MetricsMetadataTemplate,
   MetricsContainer,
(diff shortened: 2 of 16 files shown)


src/js/bugutils/minimal_pubsub.js
```diff
@@ -80,12 +80,20 @@ define(['underscore', 'backbone',
...
+          defer.done(function () {
+            context.done.call(context.context, response);
           });
           defer.resolve(response);
           return defer.promise();
-        }});
+        }
+      };
+      //should this be true for all cases?
+      if (options.hardenedApi){
```
romanchyla: pls change to:

return new Facade({'request': 'call'}, this)

src/js/widgets/metrics/widget.js
```diff
@@ -337,6 +335,14 @@ define([
       Marionette.bindEntityEvents(this, this.view, Marionette.getOption(this, "viewEvents"));
     },
 
+    activate : function(beehive){
+      _.bindAll(this, "setCurrentQuery", "processResponse");
+      this.pubsub = beehive.Services.get('PubSub');
```
romanchyla: I have started removing this.pubsub, the current pattern of activate() is

```
this.setBeeHive(beehive)
var pubsub = beehive.getPubSub();
pubsub.subscribe()
```

and then, if one needs to work with pubsub, you can call this.getPubSub()

this change is very new (I pushed it yesterday); wheneve you have a chance, pls convert the this.pubsub and this.beehive to the new 'style'

src/js/widgets/metrics/widget.js
```diff
@@ -367,36 +373,31 @@ define([
...
+      var request =  new ApiRequest({
+        target: ApiTargets.SERVICE_METRICS,
+        query: new ApiQuery({"bibcodes" : bibcodes}),
+        options : {
+          type : "POST",
+          contentType : "application/json",
+          done : done
+        }
+    });
+      this.beehive.getService("Api").request(request);
```
romanchyla: I'm (still) unconvinced here - have to really think about it, few problems here

1. no error handling
2. one has to know/specify options - instead of widgets just sending apirequest
3. the synchronization between widgets?

I get it that it is simpler, but we need to make it simpler AND at the same time robust

pls use the old mechanism of EXECUTE_QUERY for now, I have to explore consequences

btw: for this to work, you had to expose the Api.request to the widgets; more things made availalbe to the widgets means more troubles (in general; since they will not be isolated)

test/mocha/js/widgets/metrics_widget.spec.js
```diff
@@ -1070,7 +1067,7 @@ define([
             return testData;
           }
         }
-      }))({verbose: false});
+      }))({verbose: false, hardenedApi : true});
```
romanchyla: i don't see any change in js/services/api.js - is it working for you in the website? perhaps forgotten to add that file to the commit?

src/js/widgets/metrics/widget.js
adsabs-beer-user: hmm ok I understand what you're saying, I will change it back.

One thing that I didn't like about the original implementation (the reason
I changed it) is that processResponse was just checking whether the
response was an instanceof ApiResponse vs jsonResponse to decide whether it
was from solr or metrics endpoint, what is a better way to distinguish them?

src/js/widgets/metrics/widget.js
aholachek: so now, instead of this.pubsub, each function will have to call this.getBeeHive().getPubSub(), correct?

are these changes in master or IAU? They don't seem to be in the IAU branch. What should I do?

romanchyla: For IAU we have to use the old this.pubsub, yes, painful...

PR #504, opened 2015-07-21
activate dropdowns on hover

aholachek:
ok, true, this is in no way vital for the IAU and we can wait if necessary, but I think this improves the user experience a bit


romanchyla:
it seems i can't change the target branch; it has to be done when crating the PR - can you please make a new PR and select the IAU branch? (i'm assuming you have based your commit on IAU branch)

adsabs-beer-user:
I made a new pull request

PR #505, opened 2015-07-21
activate dropdowns on hover

aholachek:


romanchyla:
Alex, i forgot one thing: because we are adding a new library, it is important to update the Gruntfile - to generate the path to the minified version of the library (around line 745)

if the library is not minified, we have to minify it ourselves - look at exec:minify

btw: if you added some new libs in the past two months, this should be done for them too

aholachek:
I decided that this could be distracting for people, so I am closing this pull request for now

PR #511, opened 2015-07-22
updated getPageVal function

aholachek:
and added scroll top when paging through results
updated tests

Hi Roman,
I think the two problems were: the getPageVal function was not returning the right number in all cases, and the start value was being updated based on solr responses rather than what it actually was (so if you needed to request data 2x from solr to go from records 0-20, for instance, it would update your start value to "10" when it got the second batch of results back, which caused a never ending loop).

Can you take a look and tell me if this makes sense?
I wonder if it wouldn't be more simple to forget about keeping a hidden collection and all the extra functions that come with that, and just re-request the data each time, since we are now caching solr requests?


romanchyla:
i've merged this into IAU

Issue #501, opened 2015-07-20
Links to "view this list in a search results page" fail for A&A bibcodes


aaccomazzi:
The ampersand must be messing BB up.  For example:

this page renders properly: https://ui.adsabs.harvard.edu/#abs/2000A&AS..143....9W/citations

but when clicking the "view this list..." link, the url changes but the page contents don't: https://ui.adsabs.harvard.edu/#search/q=citations(bibcode:2000A&AS..143....9W)

aaccomazzi:
I see the discussion in #513, is there a resolution on how to proceed?  This bug is pretty nasty as it affects all bibcodes with "&" in them so we need to close it.

aholachek:
fixed by #557

PR #513, opened 2015-07-22
added regex for splitting queries

aholachek:
this addresses issue #501

### DIFF
--- a/src/js/components/multi_params.js
+++ b/src/js/components/multi_params.js
@@ -193,7 +193,11 @@ define(['backbone', 'underscore', 'jquery'], function(Backbone, _, $) {
       if (!resp.trim()) {
         return attrs;
       }
-      var hashes = resp.slice(resp.indexOf('?') + 1).split('&');
+      //there can be ampersands in the q parameter
+      //now using a positive lookahead that expects an = before any " characters,
+      //ex: "q=bibstem:"a&a"&sort=date+desc" will now be split into ["q=bibstem:"a&a"", "sort=date+desc"]
+
+      var hashes = resp.slice(resp.indexOf('?') + 1).split(/&(?=[^"]+\=)/);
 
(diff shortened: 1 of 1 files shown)


src/js/components/multi_params.js
```diff
@@ -193,7 +193,11 @@ define(['backbone', 'underscore', 'jquery'], function(Backbone, _, $) {
       if (!resp.trim()) {
         return attrs;
       }
-      var hashes = resp.slice(resp.indexOf('?') + 1).split('&');
+      //there can be ampersands in the q parameter
+      //now using a positive lookahead that expects an = before any " characters,
+      //ex: "q=bibstem:"a&a"&sort=date+desc" will now be split into ["q=bibstem:"a&a"", "sort=date+desc"]
+
+      var hashes = resp.slice(resp.indexOf('?') + 1).split(/&(?=[^"]+\=)/);
```
romanchyla: while this solves the given bug (it relies on kinda arbitrary assumption) 

the original problem is that the query must be wrongly encoded, look at this, the multi_params can handle the '&' well

```
q = new ApiQuery({title: 'a&a'});
q.url()
"title=a%26a"
q.load(q.url()).url()
"title=a%26a"
```

romanchyla: which means: we need to fix the root cause of the problem (probably some widget is creating a wrong query)

aholachek:
but if it's a widget problem, why does this url not work:

https://ui.adsabs.harvard.edu/#search/q=bibstem%3A%22a%26a%22&sort=date+desc

romanchyla:
answer to both questions lies on line 299 ;-)

https://github.com/adsabs/bumblebee/blob/master/src/js/apps/discovery/router.js#L299

romanchyla:
@aholachek gentle nudge

romanchyla:
solved in #557

PR #516, opened 2015-07-23
Fix ghostery

romanchyla:


aaccomazzi:
Why do we keep reducing the default number of bibcodes used for metrics? Today I demoed this to a scientist who has over 900 records, and he got all his metrics at once, which I think I'd how it should be. Most people will find it a nuisance to change the number (if they even notice it).

romanchyla:
the 300 was the original default and you yourself said whatever we used before was fine to revert to

but there is a more important reason: 1000 bibcodes means much higher risk of unresponsive or even crashed service; the metrics is at least O(N^2) - so if  everybody loads 1000bibcodes by default, then it is a high load (for nothing) 

btw: we didn't switch to the query call yet, so it is *actually* collecting and sending 1000 bibcodes, no kidding - 148KB for every request

PR #527, opened 2015-07-29
added 3 forms for front page, + tests

aholachek:


romanchyla:
some text is not visible:

image

romanchyla:
i did run grunt less

PR #534, opened 2015-07-31
standardized abstract and abs

aholachek:
this in response to Alberto's email this morning


romanchyla:
I'd like to see if we can find a better name than 'abs' before pushing it (btw: Alberto's feedback, as I understood it, was about the new forms - not the autocomplete)

PR #546, opened 2015-08-08
Public library

aholachek:
Sorry about the Gruntfile re-formatting, I have no idea when that happened.

One issue is that it seems that everything is limited to 100 bibcodes when you make a request to solr right now, is that how it's going to be indefinitely? 

I wasn't really sure what the best strategy for keeping the library metadata collection in sync was, but for now the library controller has a collection that it updates every time a library is deleted, or added, or altered by adding or removing a record or changing permissions, and it sends an event to listening widgets. This reduces requests when compared to the original idea, which is every time a widget wants metadata about the libraries, it requests the data.


src/js/apps/discovery/navigator.js
```diff
@@ -85,6 +89,11 @@ define([
           this.route = '#index/' + queryUpdater.clean(q).url();
         });
 
+        this.set("404", function(){
+          $("#body-template-container").html(ErrorTemplate());
```
romanchyla: should be, i think, NotFoundTemplate

also, if we have '404', should we have '505' too?

src/js/components/library_controller.js
```diff
@@ -146,7 +159,7 @@ define([
 
         if (options.bibcodes == "all"){
 
-          var limit = options.limit || 2000;
+          var limit = options.limit || 1000;
```
romanchyla: so there should also be some hardcap limit (somewhere), like Math.min(1000, options.limit || 1000)

and to answer your question, yes - we are going to limit the number of docs returned by api (max 100), so we'll have to change bbb to make multiple paging requests

romanchyla: so far these limits are applied to staging api only

src/js/widgets/libraries_all/templates/libraries-list-container.html
```diff
@@ -5,6 +5,9 @@
         <td>
             <!--page num -->
         </td>
+        <td>
```
romanchyla: some jquery is going to put values here? can it be a part of the displayed mode? (and update the model instead?)

src/js/widgets/libraries_all/views/view_all_libraries.js
```diff
@@ -2,96 +2,20 @@ define([
   "marionette",
   "hbs!../templates/libraries-list-container",
   "hbs!../templates/library-item",
-  "hbs!../templates/no-libraries"
+  "hbs!../templates/no-libraries",
```
romanchyla: if we use the optimized concatenated js, will those relative links still work as they should?

romanchyla:
Hi @jonnybazookatone could you please help me review this branch? I'm looking into it too, but you know the backend and so I think it could be great if you take a look, paying attention to the api calls.
Thanks,

jonnybazookatone:
Sure. I will take a look in the next days.

src/js/widgets/libraries_all/views/view_all_libraries.js
aholachek: yes

src/js/components/library_controller.js
```diff
@@ -238,32 +233,36 @@ define([
        * requires id, name, description
        */
```
jonnybazookatone: General nit-pick and a question for curiosity. The comment is meant to reflect what the end point wants, or what the end point returns? 

What is the common approach in the front-end to validate the input. I probably didn't get to that part of the code yet, but do you validate what's in data before you send it to the end point? Just curious about the common practice.

jonnybazookatone:
I went through the end points in both the source javascript, and tests. Everything looks fine. I also went through the two main widgets and the corresponding templates, and mixins. I have nothing to say that should block this PR. Let me know if either of you want me to check something else.

PR #549, opened 2015-08-11
fixed grant facet; reducted redundant calls to refreshLogicTooltip

aholachek:
The timeline showed that we were calling the facet method "refreshLogicTooltip" ~1200x per search! It was listening to way too many events. Now it only happens 1x per widget collection refresh, and the timeline view says that it increases the frame rate, it seems slight but noticeable (but maybe that's just because I want it to be noticeable).

I also added a condition to the grant facet function that I copied from the main "handleLogicalSelection" function

### DIFF
--- a/src/js/widgets/facet/container_view.js
+++ b/src/js/widgets/facet/container_view.js
@@ -36,21 +36,14 @@ define(['backbone', 'marionette',
             throw new Error('logicOptions should be null or an object with single/multiple keys and arrays of strings inside');
           }
 
-          this.on('all', function(ev, info) {
+          this.on('all', function (ev, info) {
             if (ev.indexOf('itemClicked') > -1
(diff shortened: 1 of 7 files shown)


romanchyla:
I think it is faster. (though we still listen to all events - we should probably go through the code and replace those onAll() handlers with more specific ones - because this function is still called 1200 times per search

1200 times you say? that seems a lot - is it making sense (from one listener?)

not sure it will still work with multiple (n>3) levels - probably not, because the ev == render:collection will miss the deep nested

aholachek:
To reproduce, you can put a console.count() statement in the function then search something big like "star"

romanchyla:
What I meant: does it make sense it should be called 1200 times? I'm not
questioning that it is actually called that many times.

PR #552, opened 2015-08-13
Requirejs concat

aholachek:
On my computer right now, which is admittedly moving pretty slowly, it takes ~2000 ms for current bumblebee just to load widgets and associated files.

the concatenated version does it in  ~140 ms

I needed to make two major changes to the gruntfile: adding the function that grabs the list of bbb dependencies from discovery.config and assigns it to the "include" parameter for the require release task, and grunt "string replace" to have the production index.html access a file called "concatenated bumblebee" instead of discovery.config as the data-main attribute for requirejs.

The one issue is the query builder plugin is still causing the optimized file to crash, so I've commented it out to make the pull request so it can be looked at, but I will keep trying to fix it now. I'll add the error message in a second.


vsudilov:
Nice work! Nearly there!

:shipit:

PR #556, opened 2015-08-13
automated concatenation of bumblebee

aholachek:
Hi Roman,

I fixed the querybuilder thing (it was the shim that had the "exports" parameter which doesn't  need to be there, I removed it)


package.json
```diff
@@ -27,22 +29,21 @@
...
     "grunt-env": "~0.4.1",
     "grunt-exec": "~0.4.5",
     "grunt-express-server": "~0.4.13",
+    "grunt-hash-required": "^0.1.5",
     "grunt-include-bootstrap": "~0.1.1",
     "grunt-install-dependencies": "~0.1.0",
+    "grunt-mocha": "^0.4.11",
     "grunt-mocha-phantomjs": "~0.4.3",
     "grunt-processhtml": "~0.2.0",
-    "needle": "~0.7.1",
```
romanchyla: just a side note; your editor has some aggressive re-formatting setup

src/index.html
```diff
@@ -28,49 +28,29 @@
...
             </div>
-
-
-        </div>
-
+            <span class="loading-screen__loading">loading...</span>
         </div>
 </div>
 
-<script>
```
romanchyla: him, if this place is vacant, does it mean we can bring back the silly nerdy pacman? or you have some other evil plans?

aholachek: i think since the load time is so much faster, it won't be necessary? maybe that's wishful thinking. I'm just showing the ads logo right now

romanchyla: it may be not as fast on slow line

PR #559, opened 2015-08-17
cleaning up variables and mixins, fixing small things

aholachek:


aholachek:
Roman is it all right if I merge this into IAU after the tests pass? it's fixing the background color for the search results page which seems to have gone missing at some point

PR #565, opened 2015-08-19
added version number

aholachek:
i also got rid of those two example apps, i think they make things confusing for people perusing the files


romanchyla:
merged

PR #569, opened 2015-08-21
Life cycle

romanchyla:
This refactoringn is not yet finished (tests are missing); I'll have to return to it - but please take a look if you have a chance.

It works as it should - the page managers/widgets are created on the fly (but I think they are not being destroyed - the counter is 1step too high)

There is also a problem with the abstract view: the search bar is there, but for some mysterious reason, the it is hidden (i was not able to find the place - a template? a jquery code?)


romanchyla:
OK, I've solved the 'stale subscribers' problem

aholachek:
Hi Roman,

I'm still going through these changes and trying to understand them. It seems to load a second or two faster than the non-minified current master, though I think the numbers from the Chrome timeline view are not exact, that's still promising.

A few questions:
-Right now the back button to the search page doesn't work, is that a bug?
-The search bar widget disappears now when you click search, at least on the landing page, which is kind of unnerving.
-BBB feels slower since the abstracts are no longer pre-fetched and since I assume when it works, the back button will have to re-fetch the data to show the earlier page. Are there ways around this?

romanchyla:
Hi Alex,

- the back button (and e.g. going from user preferences back to list of
results) does not work because the navigator will have to be updated; to
assemble the page before the signal is called; so that the widgets have
chance to respond - it is something i should fix (as I've done for all the
other pages)

- yes, the search bar disappears, and it bothered me too - but not enough
to dig what it was (but now I repent and I'll....:)))

- and yes, it feels slower - and I don't like it either - maybe after you
finish studying the changes, you will get some ideas which direction this
should go?

Thanks,

  roman

aholachek:
As you can probably tell I was trying to find ui problems to delay the inevitable "having to understand what is actually going on" part

romanchyla:
and as you can probably tell, i did let the first two problems slip away
for a similar reason

src/js/components/application.js
```diff
@@ -485,78 +593,247 @@ define([
...
+
+      var pubsub = this.getService('PubSub');
+      var existingSubscribers = _.keys(pubsub._issuedKeys);
+
+      if ('activate' in instance) {
+        if (this.debug) {console.log('application: ' + symbolicName + '.activate(beehive)')}
+        children = instance.activate(hardenedBee);
+      }
+
+      var newSubscribers = _.without(_.keys(pubsub._issuedKeys), _.keys(pubsub._issuedKeys));
```
aholachek: what is the purpose of this line? Is it doing anything other than making a copy of the array?

src/js/components/application.js
```diff
@@ -485,78 +593,247 @@ define([
...
+
+      // we'll monitor all new pubsub instances (created by the widget) - we don't want to rely
+      // on widgets to do the right thing (and tells us what children they made)
+
+      var pubsub = this.getService('PubSub');
+      var existingSubscribers = _.keys(pubsub._issuedKeys);
+
+      if ('activate' in instance) {
+        if (this.debug) {console.log('application: ' + symbolicName + '.activate(beehive)')}
+        children = instance.activate(hardenedBee);
```
aholachek: does the "activate" method ever return a list of child widgets?

src/js/components/application.js
```diff
@@ -485,78 +593,247 @@ define([
...
+      }
+
+      var childNames = [];
+      if (children) {
+        childNames = this._registerBarbarianChildren(symbolicName, children);
+      }
+
+      if (illegitimateChildren){
+        _.each(illegitimateChildren, function(childKey) {
+          if(this.__barbarianRegistry[childKey]) // already declared
```
aholachek: For the widgets on the search page and the abstract page, since they will be presumably used many times and we will want to keep their state for the back button, and maybe pre-render them for the abstract page, could we just *not* delete them once they are instantiated? Then use this new delete/re-instantiate strategy for all other, less heavily used widgets?

src/js/components/application.js
romanchyla: that finds the new subscribers (those, that got created between lines 644-646)

src/js/components/application.js
romanchyla: yes, in some cases - graph widget maybe? i'm sure it does in at least 2

src/js/components/application.js
romanchyla: yes, but "how to", what would be the right strategy?

src/js/components/application.js
aholachek: I still don't quite understand: isn't _.keys(pubsub._issuedKeys) just repeated 2x? Won't it always yield the same array? Also, I don't think you can use the _.without function like this, it takes individual arguments, not an array-- test it by using
_.without([1,3,4], [1,3])

Sorry if there's something huge I'm missing

src/js/components/application.js
aholachek: Well, the probably dumb way I got this working was to add an extra value in .__barbarianInstances, "preventKill", that is passed in by the controller doing the _getWidget call, and causes killBarbarian to simply return--and adding this flag to the ResultsPageController and AbstractPageController.

This solution is not perfect because it's holding on to widgets like the author network widget that we probably don't want to keep around. In addition the back button is still not as fast as before because I think the page manager is still having to insert the widgets.

Is this roughly something that could work though? Maybe including a flag on a per-widget basis, rather than page-manager basis, that overrides the logic in killBarbarian? (these names are very silly by the way, I like them)

src/js/components/application.js
romanchyla: oups, sorry! i've - and that is shameful from my side - assumed that you were asking about the purpose of the code. And I saw only first half of the line on my laptop. 

You are of course right, should have been:

_.without(_.keys(pubsub._issuedKeys), existingSubscribers);

btw: this code is not causing any problems because I've deactivated the functionality related to it (but I'll try to activate it again)

src/js/components/application.js
romanchyla: try to think about the counter - notice that the widget gets killed only when it is not referenced (its counter reaches zero) - perhaps this mechanism would do the job

aholachek: So, increment the counter +1 if it's a widget that we want to keep around, so it is never de-referenced? But where does this information about the longevity of the widget live? In the widget's constructor itself?

romanchyla:
hi Alex, the two issues are fixed (and i've introduced small changes such as getWidgetRefCount() method - and also the discovery mediator using it to decide whether a search needs to be re-executed)

PR #592, opened 2015-09-12
added export to png/save as pdf option for metrics

aholachek:


aholachek:
this works nicely in chrome but the image formatting is off in firefox so I'm closing the pull request

PR #604, opened 2015-09-29
Iau

aholachek:


romanchyla:
before I start: this is the first time I'm seeing PL in action and I REALLY LIKE IT! It is easy to use and in general it makes sense. Good job!

OK, now back to the review: it is very difficult to review them all - i'm going first through functionality, here is what i found (and I'm not sure whether the problem is in master or in the merged branch):

- export (shows just the first paper, not all X)

- when viewing metrics, the url changes from http://localhost:8000/#search/fq=%7B!type%3Daqp+v%3D%24fq_keyword_facet%7D&fq_keyword_facet=keyword_facet%3Aism%5C+dust%5C+extinction&q=(author%3A%22Bernard%2C+J.+-P.%22+AND+year%3A2002-2010)&sort=classic_factor+desc   into    http://localhost:8000/#

- add papers to library (there is some tag escape error in the display); this happens only when I have zero PLs (btw: the blue PL widget is taking space - it pushes the histogram down; maybe it could be merged with the 'selected papers' widget)

- in results grapsh, selecting region is black 
image

- and the (X) button doesn't work

src/js/page_managers/controller.js
```diff
@@ -116,7 +116,12 @@ define([
           if (widget.disAssemble)
             widget.disAssemble();
           app.returnWidget(widgetName);
-          $(this.widgetDoms[widgetName]).empty();
```
romanchyla: i'm failing to see the reason for this; page-managers always rebuild stuff

btw: better to use app.getWidgetRefCount() -- the _methods are private and should not be used outside of the app (nad yes, i was mysefl inconsistent here in the case of _getWidget())

aholachek: if we empty, we undelegate all widget events
which means the next time the widgets (that are still alive) are inserted into the dom, none of their events will work

PR #632, opened 2015-10-14
added urls for metrics + visualizations (not export for now since tha…

aholachek:
…t would need an extra argument)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -593,7 +593,7 @@ module.exports = function(grunt) {
           modulePattern : "../../js/(.*)",
           customModuleThreshold: {
             "widgets/alerts/widget.js" : 73,
-            "apps/discovery/router.js": 34,
+            "apps/discovery/router.js": 32,
             "widgets/facet/graph-facet/h_index_graph.js":2,
             "widgets/facet/graph-facet/year_graph.js":2,
             "wraps/graph_tabs.js":5,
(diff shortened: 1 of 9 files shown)


Gruntfile.js
```diff
@@ -593,7 +593,7 @@ module.exports = function(grunt) {
           modulePattern : "../../js/(.*)",
           customModuleThreshold: {
             "widgets/alerts/widget.js" : 73,
-            "apps/discovery/router.js": 34,
+            "apps/discovery/router.js": 32,
```
romanchyla: in this way, little by litte, the coverage will decrease - and the showResultsPageWidgetWithUniqueUrl is nicely unit-testable

src/js/apps/discovery/router.js
```diff
@@ -90,7 +90,9 @@ define([
 
       },
 
-      search: function (query) {
+      search: function (query, widgetName) {
```
romanchyla: could this logic be moved into the navigator? i.e. let handle router always call navigate? (i guess I'm not understanding the necessity of having the hardcoded list of widgets here, and as well in the navigator)

aholachek: The list is only used so that the router can call an error if the person requests a widget that doesn't exist or uses the wrong name -- I can get rid of it, and then it will become a navigator error, is that better?

romanchyla: calling NAVIGATE with "show-" + widgetName seems very reasonable to me; because when it is not there, the error gets raised -- but the logic under review here is different, it happens only for results-page (and silently ignores some other cases; that is not better)

PR #655, opened 2015-10-29
fixed public libraries (still needs tests), merged in last of IAU com…

aholachek:
…mits, fixed 'modern search' routing issue

### DIFF
--- a/src/discovery.config.js
+++ b/src/discovery.config.js
@@ -40,7 +40,7 @@ require.config({
           FeedbackMediator: 'js/wraps/discovery_mediator',
           QueryMediator: 'js/components/query_mediator',
           Diagnostics: 'js/bugutils/diagnostics',
-          AlertsController: 'js/components/alerts_mediator',
+          AlertsController: 'js/wraps/alerts_mediator',
           Orcid: 'js/modules/orcid/module'
         },
         services: {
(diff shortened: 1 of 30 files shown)


romanchyla:
Alex, what is the reason for applying those changes manually?

aholachek:
At first, I tried to rebase IAU on top of current master (like I did for
the initial merge) but for reasons I couldn't understand, it was showing
conflicts in many files that should not have been touched in the most
recent IAU commits (which were exclusively for alerts).

I don't think it was a good idea to go through the most recent commits and
manually apply the changes, though, since for a few files the formatting
changes meant it was impossible to see what had actually changed, so it was
pretty error prone and I missed a few things that I've been fixing today.
Do you have any advice what I should have done instead?

romanchyla:
The basic advice is not to commit to two branches (IAU and master) the same stuff. The IAU has to be decomissioned (I think the changes from there were already applied to master); next is to keep the development separated from production. 

I thought that was your plan - to start using the new model (that has to be done before proceeding with more development)

If it happens that you see many conflicts (for something that shouldbe clean), let me know.

aholachek:
Yes, I have not done anything on IAU since you added the alerts so that we
could show the site-wide alert. That only touched the alerts_mediator,
alerts_widget, and user component, I think. Which is why I didn't
understand why git was showing conflicts in other files.

I tried to do:
git checkout IAU
git rebase upstream/master

and it showed me tons of conflicts, so something I did was clearly wrong
but I don't know what. Anyway, after yesterday's messy merge of the
site-wide alerts, IAU can be safely deleted.

PR #686, opened 2015-11-23
added bigquery capabilities to bibcode form + libraries

aholachek:


package.json
```diff
@@ -42,10 +42,9 @@
     "grunt-processhtml": "~0.2.0",
     "grunt-sass": "^1.0.0",
     "grunt-saucelabs": "~8.6.1",
+    "grunt-string-replace": "^1.2.0",
     "intern": "~3.0.0",
     "load-grunt-tasks": "^3.2.0",
-    "grunt-saucelabs": "~8.6.1",
```
romanchyla: why removing?

src/js/apps/discovery/navigator.js
```diff
@@ -411,7 +411,15 @@ define([
           app.getObject('MasterPageManager').show('SearchPage',
             searchPageAlwaysVisible);
           var q = app.getObject('AppStorage').getCurrentQuery();
-          this.route = '#search/' + queryUpdater.clean(q).url();
+          var route = '#search/' + queryUpdater.clean(q).url();
+
+          //taking care of inserting bigquery key here, not sure if right place
```
romanchyla: i noticed that reloading the url with 'qid' doesn't work - '__qid' will work, i think you should settle on 'qid'

src/js/components/query_mediator.js
```diff
@@ -92,6 +92,48 @@ define(['underscore',
...
 
+      getQueryAndStartSearchCycle : function(apiQuery, senderKey){
+
+        var that = this;
+
+        // checking if it's a new big query
+        if (apiQuery.get("__bigquery")) {
+
+          var query = new ApiQuery({
+            q  : '*:*',
```
romanchyla: this is not generic enough

You should reuse all parameters (i.e. not replace 'q' with '*:*') and once you get the 'qid' back, just create a new query

new ApiQuery({'qid': qid})

btw: since this is a query-mediator, you can (and probably should) call api.request().done().fail() - but maybe you already do that (or it is done)

src/js/components/query_mediator.js
```diff
@@ -169,14 +211,21 @@ define(['underscore',
...
-
+        if (cycle.running) return; // safety barrier
         if (_.isEmpty(cycle.waiting)) return;
         if (!this.hasBeeHive()) return;
 
+        // for altering widget queries
+        // from regular solr requests to execute_query requests
+        // if bigquery is being used
+        function makeBigQuery(request){
+          var qid = request.get("query").get("__qid")[0];
```
romanchyla: use 'qid'

src/js/components/query_mediator.js
```diff
@@ -92,6 +92,48 @@ define(['underscore',
...
+          })
+
+          var request = new ApiRequest({
+            target : ApiTargets.MYADS_STORAGE + "/query",
+            query : query,
+            options : {
+              type : "POST",
+              contentType : "application/json",
+              done : function(response){
+                that.startSearchCycle( new ApiQuery({q : query.get("q"), "__qid" : response.qid }), senderKey)
```
romanchyla: i see: but it should be only new ApiQuery({'qid': response.qid})

also, the error handling is somewhat missing in functionality :)

src/js/components/query_mediator.js
```diff
@@ -92,6 +92,48 @@ define(['underscore',
...
+              }
+            }
+          });
+
+          //it needs to be a bigquery, we need to formulate the query
+          this.getBeeHive().getService("Api").request(request);
+
+        }
+        // pre-existing big query--maybe from faceting or url load
+        else if (apiQuery.get("__qid")){
```
romanchyla: if you already have 'qid' - you have the logic in _executeRequest  (actually, much better place would be executeRequest)

package.json
aholachek: it's still there on line 44, not sure why the list was reordered

PR #703, opened 2015-12-09
Fixed load profile errors because of missing name/surname

romanchyla:


src/js/bugutils/diagnostics.js
```diff
@@ -36,7 +36,7 @@ define([
       this.setBeeHive(beehive);
     },
 
-    getFirstDoc: function(queryString, options) {
+    getFirstDoc: function(queryStrirg, options) {
```
aholachek: hi @romanchyla is this a typo?

romanchyla: yes, i'll fix it - and commit, unless there is some other problem

PR #722, opened 2015-12-23
show feedback to user instead of raising an error

aholachek:


aholachek:
hi @romanchyla, can you take a look at this and tell me if you think it will fix the loading problem? It seems like it would, however I could not reproduce the exact orcid/requirejs issue on localhost when I changed the orcid api info in the discovery.vars as you suggested, so I'm not sure if it's fixed

PR #723, opened 2015-12-24
Bulk orcid add

aholachek:
not ready to be merged, just in case Roman can take a look

### DIFF
--- a/src/js/modules/orcid/orcid_api.js
+++ b/src/js/modules/orcid/orcid_api.js
@@ -153,7 +153,6 @@ define([
             (targetRoute || '/#/user/orcid'))
         });
         //make sure to redirect to the proper page after sign in
-        this.getPubSub().publish(this.getPubSub().ORCID_ANNOUNCEMENT, "login");
         var currentPage = this.getBeeHive().getService("HistoryManager").getCurrentNav();
         this.getBeeHive().getObject("AppStorage").setStashedNav(currentPage);
 
@@ -220,7 +219,8 @@ define([
        */
       signOut: function () {
         this.saveAccessData(null);
-        this.getPubSub().publish(this.getPubSub().ORCID_ANNOUNCEMENT, "logout");
+        //we just authenticated. inform other widgets:
+        this.getPubSub().publish(this.getPubSub().ORCID_ANNOUNCEMENT, this.ORCID_SIGNED_OUT);
       },
 
       hasExchangeCode: function (searchString) {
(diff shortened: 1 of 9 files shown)


src/js/widgets/library_individual/widget.js
```diff
@@ -69,6 +70,26 @@ define([
...
+        this.headerModel.set("orcidSignedIn", orcidLoggedIn);
+        //views need to be updated manually, they do not listen to change events
+        this.syncHeader(this.libraryCollection.toJSON());
+      },
+
+      onOrcidAnnouncement : function(msg){
+        //headerview automatically listens to change event on orcidSignedIn
+
+        var orcidApi = this.getBeeHive().getService("OrcidApi");
+        if (msg === orcidApi.ORCID_SIGNED_OUT){
```
romanchyla: don't do this, it is an identity check (it works here only because the strings are immutable, but it is not a good practice)

src/js/modules/orcid/orcid_api.js
```diff
@@ -1060,6 +1060,109 @@ define([
...
+                  var successfullyAdded = 0, failAdded = [];
+                  //fall back to adding one-by-one
+                  toAdd.forEach(function(record){
+                    that.updateOrcid("add", record)
+                        .done(function(){
+                          successfullyAdded+=1;
+                          checkAddDone();
+                        })
+                        .fail(function(){
+                          failAdded.push(record);
```
romanchyla: you should keep track of recs that failed (ie: their bibcodes)

romanchyla: oh, but since this method is a part of orcid_api; it should definitely return the ids of updated as well as failed docs

src/js/modules/orcid/orcid_api.js
```diff
@@ -1075,6 +1075,109 @@ define([
...
+            }
+            else {
+              //update it
+              toUpdate.push(record);
+              checkDone();
+            }
+          })
+        });
+
+        //attempt to bulk add then bulk update
```
romanchyla: it would be nice to add a timeout as well; because the checks may never finish - actually, the getRecordInfo() should have a timeout and fail (so that would be cleaner)

romanchyla:
hi alex, it looks good, try to add tests though - especially for the batchUpdate(), that should be tested in a similar fashion to addWorks(), updateWorks() -- those tests are already in the orcid_api.spec.js -- so you can use them as a blueprint

aholachek:
ok, thank you, i will look at those tests for examples

PR #756, opened 2016-01-14
SIMBAD facet attempt

ehenneken:
@romanchyla I have implemented changes you suggested and added caching. I don't know if I have implemented caching in the right way. I now check whether the incoming response is a SolrResponse, to determine what to do. I have no idea how to make a test for the object facet. Now the code doesn't seem to get executed anymore :( It looks like I'll need to guidance again.


ehenneken:
@romanchyla I realized that in this PR, the `SERVICE_OBJECTS` definition is missing in `api_targets.js`

ehenneken:
I'm wondering if I need to do something like

    facetCollection.forEach(updater)

rather than

    this.collection.each(updater);

just a hunch, not based on any knowledge :-)

src/js/wraps/object_facet.js
```diff
@@ -0,0 +1,108 @@
...
+          var vv = v.split('/');
+          return vv[vv.length - 1];
+        }
+      ]
+    });
+	widget._getNewCache = function(options) {
+	  	return new Cache(_.extend({
+	  		'maximumSize': 150,
+			'expiresAfterWrite':60*30 // 30 mins
+	  	}, _.isObject(options) ? options : {}));
```
romanchyla: you can remove the `_.isObject()` call - it is failing as you didn't import 'underscore' module (that is likely the reason why the thing doesn't execute anymore)

which reminds me: did I show you how to create a unittest? :)

i'd recommend creating a spec (follow this: https://github.com/adsabs/bumblebee/blob/master/test/mocha/js/widgets/hello_world_widget.spec.js) - that will quickly show you the errors

ehenneken: I looked at the unittests, but didn't understand where to start and how to test... but I'll look at the "Hello World" example. Thanks for the help!

src/js/wraps/object_facet.js
```diff
@@ -0,0 +1,108 @@
...
+			//
+	        // For non-empty lists of facet strings, translate the SIMBAD ids
+			if (ident.length > 0) {
+				var identifiers = ident.map(function(v){
+					var vv = v.split("/");
+					return vv[vv.length-1]
+				});
+				this.getSIMBADobjects(identifiers);
+			};
+			var updater = function(model) {
```
romanchyla: keep in mind the asynchronous nature of ajax - when updater gets executed, you may still lack the data

so, i'd say - call the processFacets (L:81) but move the updater (L:68-80) into the other branch (L:85) and update any current facet there

something like: 

```
for each facet:
  if facet.get('objId') === facet.get('title') && v = self._cache.getSync(facet.get('objId')))
    facet.set('title', v)
```

ehenneken: My implementation of the cache was correct?

romanchyla: yes (besides the use of 'unimported module')

src/js/wraps/object_facet.js
romanchyla: for the start, create a new file `tests/widgets/object_facet.spec.js` and `test/mocha/sandbox.js`

the .spec file can be a copy of the hello world spec

the sandbox file can look like

```bumblebee master → cat test/mocha/sandbox.js 
define([], function() {
  var tests  = [
    '/widgets/object_facet.spec.js'
  ];
  return tests;
});
```

and then you can test it by `grunt test:sanbox`

or in browser: http://localhost:8000/test/mocha/tests.html?bbbSuite=sandbox

ehenneken:
@romanchyla Two observations regarding the unittest for the SIMBAD facet:

1. I had to re-insert the `subscribe` (I did it as `minsub.subscribeOnce`, line 130), otherwise the results of the micro service were never processed by "processFacetResponse"
2. The cache is not working properly. I have a feeling there needs to be some sort of `initialize` function when the SIMBAD facet is instantiated. With the current solution (lines 81-83), the cache is not defined when a "put" is attempted on line 109. Can I add an `initialize:` to the ` var widget = FacetFactory.makeHierarchicalCheckboxFacet({...})`? Do I have to use an `extend` somewhere? Help! :-)

ehenneken:
@romanchyla Everything seems to work properly in `object_facet.spec.js`, but when I apply the same solution in the real object facet, it doesn't work :(

ehenneken:
The facet gets created successfully, the "more" link seems to work as expected, but using the facet for filtering creates problems (not with filtering, but with updating of the facet) and I don't see breadcrumbs. I received a 404 when I used a facet entry for filtering.

romanchyla:
breadcrumbs are defined outside - we have to add a filter to them to recognize this new facets (but that is a minor thing)

when you click/use the facet - what is the resulting (generated) query?

PR #781, opened 2016-02-03
simplifying updateDatabase function to prevent timeout bug

aholachek:
hi @romanchyla can you take a look at this and tell me if it seems like an appropriate fix for the problem?

### DIFF
--- a/src/js/modules/orcid/orcid_api.js
+++ b/src/js/modules/orcid/orcid_api.js
@@ -734,7 +734,7 @@ define([
         var options = {
           type: 'GET',
           url: url,
-          cache: this.pending ? true : false, // true = do not generate _ parameters (let browser cache responses)
+          cache: this.updateDatabasePromise ? true : false, // true = do not generate _ parameters (let browser cache responses)
           timeout: this.orcidApiTimeout,
           done: function(data) {
             result.resolve(data);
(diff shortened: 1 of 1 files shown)


src/js/modules/orcid/orcid_api.js
```diff
@@ -921,23 +924,24 @@ define([
...
-            }
-          }, 200);
-          setTimeout(function() {
-            clearInterval(tme);
-          }, 3000);
-          return whenDone.promise();
+
+        //update is currently in progress
+        if (this.updateDatabasePromise){
+          return this.updateDatabasePromise
```
romanchyla: you should return this.updateDatabasePromise.promise()

src/js/modules/orcid/orcid_api.js
```diff
@@ -921,23 +924,24 @@ define([
...
+        }
+
+        //otherwise, have to run the update and set this.updateDatabasePromise
+        this.updateDatabasePromise = whenDone;
+
+        //this function will clear the resolved whendone promise after 30 seconds,
+        //so if new requests are made, they will have to update database again
+        function clearWhenDoneAfterDelay(){
+          setTimeout(function(){
+            self.updateDatebasePromise = null;
```
romanchyla: call .reject() first, then set to null

romanchyla: oh wait, why this function? if I remember correctly, if we set the timeout parameter, then this is not necessary

aholachek: what will happen to the requests for _getRecordInfo from other item views that come back after the promise has been resolved? if we remove the deferred right after it returns, they will show errors, right?

PR #793, opened 2016-02-12
Embedapp

romanchyla:
Standalone widget + widget for the Green button

for review: do not merge yet!


test/mocha/js/widgets/green_button_widget.spec.js
```diff
@@ -0,0 +1,106 @@
...
+           ApiResponse,
+           ApiRequest,
+           ApiQuery,
+           MinimalPubSub,
+           GreenButtonWidget
+    ) {
+
+    describe("Green Button Widget (green_button_widget.spec.js)", function () {
+
+      beforeEach(function () {
```
romanchyla: hi @jonnybazookatone, please look at this

the widget is basically ready, when it receives data, it will render them (i have to plug in the actions, but they will be something like /command=deploy&version=xxxxx&application=sandbox&environment=adsws)

I've decided to render every environment independently (because that should be easier for the api)

the api just needs to return database records for every *active* service/target - there is some extra information needed, namely

 - success
 - previous_version (actually, better to supply the array of previous_versions - up to max of 20(?))

the adsdeploy model has git/release -- those two should be merged into a 'version'; and status and msg could provide 'just in time' updates about the health of the service

jonnybazookatone:
Looks good. Just to write down what we spoke about (all is also relevant to PR adsabs/ADSDeploy/issues/14):

  * there is a `/status` GET end point that returns all *deployed* application+environment, and for each of these there is a **previous_versions** list available, eg.,
```
GET /status

[{
  u'previous_versions': [u'commit-1', u'commit-2'], 
  u'status': u'success',
  u'date_last_modified': u'2016-02-12T21:09:46.938290', 
  u'application': u'adsws',
  u'environment': u'staging', 
  u'tested': True, 
  u'tag': u'v1.0.0', 
  u'version': u'v1.0.0', 
  u'commit': u'commit-3', 
  u'active': [u'commit-3'], 
  u'date_created': u'2016-02-12T21:09:46.938285', 
  u'deployed': True, 
  u'msg': None
}]
```
  the 'active' field is there as a hack at the moment, as there could be a scenario in which two deployments are 'active'. I'll open an issue for this.

  * there is a WebSocket `status` name space that returns a model update, in a list, eg. 
```
[{
  'status': None, 
  'date_last_modified': '2016-02-12T20:59:54.250776', 
  'tested': None, 
  'tag': u'v1.0.0', 
  'commit': None, 
  'deployed': True, 
  'environment': u'staging', 
  'application': u'adsws', 
  'version': u'v1.0.0', 
  'msg': None, 
  'date_created': '2016-02-12T20:59:54.248449'
}]
```
Note both `status` and `msg` are then forwarded to the UI on updates. The individual workers must be in charge of modifying these properties, as the DB Writer, webapp and UI are unaware of such modifications, which I think is the best approach.

  * the `version` is currently in the following order: `tag` then `commit`, this can be changed.

  * there is a command end point that you can use to push to RabbitMQ. Currently, it is forwarded to 'ads.deploy.before_deploy'. There is no logic infront of BeforeDeploy, but this should be included (issue also open).
```
GET /command?action=action&environment=staging&application=adsws&version=v1.0.0

200, {'msg': 'success'}
```



The following issues are also open for getting the UI running by Thursday (or seen on kanban board):
  * adsabs/ADSDeploy/issues/15
  * adsabs/ADSDeploy/issues/16
  * adsabs/ADSDeploy/issues/17
  * adsabs/ADSDeploy/issues/18

PR #816, opened 2016-02-19
Improved UI for the deployment

romanchyla:


aholachek:
Hi @romanchyla I've had a look and will merge this -- just for my own edification, what was the reason you chose to make the green button widget part of bumblebee rather than its own small standalone app?

PR #825, opened 2016-02-24
limiting xhrfields.withcredentials to subset of api targets

aholachek:


src/js/components/api_targets.js
```diff
@@ -63,22 +63,44 @@ define([
...
         }
       }
 
     };
+
+    //add credential info
+
+    //requires cross domain cookies
+    config._needsCredentials = [
+
```
jonnybazookatone: `accounts/user`? Otherwise I don't see anything missing as far as I can tell.

PR #839, opened 2016-03-02
implementing pushstate on the dev server, changing all routes to "/"

aholachek:
@aaccomazzi @romanchyla 

#### Here are the major changes

1. server configuration
     - the old file shim.html is not necessary to support non-pushstate-enabled clients so it has been removed
     - All files (libs, .js, .html, .css etc) other than index.html are now located in /assets
     - If a request comes to the dev server, it checks if the path contains "/assets"
          - if it does, the requested asset file is returned
          - if not, index.html is returned
    - this fairly simple logic will have to be added to the production server
2. changes in bumblebee
  - all click events on anchor tags are listened to by the app -- if they are relative then the router is called        with trigger({navigate : true })
   - all anchor tags in bumblebee now use proper relative urls (e.g. "/abs/bibcode") rather than hash urls.
  -  requests for images and css filepaths now have /assets/ prepended to them

#### Justification
I looked around and all implementations of pushstate-aware servers that I could find (e.g. [here](http://jaketrent.com/post/pushstate-webpack-dev-server/) and [here](https://github.com/scottcorgan/pushstate-server)) use this method rather than maintaining an intricate list of routes on the server

(note: for this to work locally you have to set pushstate : true in your discovery.vars)

(another note: the new dev server behavior right now is messing up the tests, I have to figure out how to turn this off when tests are running)


romanchyla:
Alex, If all assets are served from 'assets' then I don't see the point of renaming them - you could just adjust the rule that all assets are served from 'js' (why to make massive changes in the whole codebase?)

What I'd like to see is: a justification why some approach is better than the other; if you can't find real examples of push-state projects just explore different approaches as theoretical possibilities and write down arguments in favor/against.

please don't merge this PR - touching 1038 files is most likely not the solution we are looking for

aholachek:
Wait, so assets including images css, fonts, etc should all be served from
a folder called js?


Ok, I will write up a more detailed pro and con list.

aholachek:
Push state Options 

1. use file paths (if .js/.css/.html is detected, return that file; otherwise return index.html)
     -  pro: don’t need to touch file structure
     - con: what if a new file structure is requested? (like .woff for fonts)
          - could just broaden the rule that if it has a file path at all, it is returned as a static asset
     - con: might anyone ever try to add .html to the end of a bbb url?
     - pro/con : 404s will work for static files, but not for bbbb routes

2. everything in /assets is returned as static file, everything else is attempted to route (this is the option I chose in the pull request above)
    - pro: easy to reason about, similar organization to other web projects
    - pro: very little configuration necessary now and in the future
    - pro/con : 404s will work only for static files with filepath /assets/, and false bbb routes will not return 404s
    - con: will have to move assets (css/js/html/img) into assets file, restructuring project

3. keep a list of routes on the server
    - pro: can send back proper 404s
    - pro: might force us to be more disciplined about url structure
    - con: lots of configuration, (“duplication of logic”), every time a route is added it needs to be updated on the server
    - con: debugging might be a bit of a pain: if there is a problem it could require multiple people to coordinate updating bbb + the server

aaccomazzi:
I am not personally bothered by the potential renaming of a lots of files (solution 1) or requiring the definition of some configuration on the server (options 2 and 3).  I also don't think that there is really duplication of logic in 3, but just a fairly easy-to-solve problem with configuration (bbb and server configs need to be in sync).

What concerns me the most is the proper handling of requests for resources on our site by either users (via browsers) or applications (especially search engines).   The resources I'm thinking about are the "abstract pages" for the most part, but will eventually include things such as user pages (keyed on orcids), user libraries, and possibly other ids.  What I would like to see in those cases is a system which is capable of handling scenarios in which the pages corresponding to these resources are bookmarked, shared, used to create collections, and indexed by search engines.  This mainly boils down to properly support permalinks: the URL of a resource should continue to live on (even upon the renaming of a bibcode), or otherwise return the proper "not found" response.  I think this means the following for us:

1. Having resources identified by unique URIs (no hashes), e.g. http://ui.adsabs.harvard.edu/abs/2016arXiv160107858A
2. When a resource is renamed (change in bibcode), the client is redirected to the new one (via HTTP 301/302)
3. When a resource is deleted, return a proper HTTP 404
4. Ideally, the resource returned has machine-readable metadata embedded in it (not just a javascript app)

Up until now it was not clear to me if and how this could be done using a single page javascript app.  I think I now understand that it simply cannot be properly done, and that I'm advocating having a website rather than an application handle those requests (see e.g. http://stackoverflow.com/questions/14779190/in-a-single-page-app-what-is-the-right-way-to-deal-with-wrong-urls-404-errors).

Now I'm not suggesting we get rid of BBB and starting from scratch, but rather identifying the routes associated with resources of the kind described above and looking for a server-side solution which works in concert with BBB.

Related to https://github.com/adsabs/issues/issues/27 and https://github.com/adsabs/issues/issues/30

PR #844, opened 2016-03-04
prevent non-signed-user from accessing #user/libraries

aholachek:


romanchyla:
What happens if I try to curl the endpoint without being signed in? Will I get the Pl data?

aholachek:
No, you won't have permissions so you can't access anything.

That's why bumblebee was showing a blank template instead of showing any
real data for the user/libraries view.

PR #850, opened 2016-03-08
Absrender

aholachek:
hi @aaccomazzi, I did the demo of the server side rendering 
- requires pushstate
- handles 301 redirects 
- returns 404 when necessary
- renders metadata into body + head (using highwire tags )
- wipes the head + body if/when javascript starts running
- works with Papers 

I can show it on localhost if you want

(this is not ready to be merged)


aaccomazzi:
Big improvement over current situation IMHO.  Needs a good look from @romanchyla followed by a discussion on how to proceed:
- use node.js?
- declarative list of routes vs. defaults
Also it would be nice to be able to test this against the google crawlers.  I suppose if we commit this then we can use the urls in staging and test them against googlebot.

aholachek:
I added server side routes (it imports the routes from bbb so configuration is in only 1 place) so that proper 404s can be returned from the server

romanchyla:
I see many challenges, so better if we speak about it - btw this PR duplicates #839 - the discussion should happen there or here (but not in both). And tokens should not be included in the code, even if it is testing

PR #859, opened 2016-03-15
overriding require.js waitseconds at very beginning

aholachek:
This hack is suggested (sort of) here: https://github.com/yeoman/yeoman/issues/1051 for  bundles compiled with r.js.

It definitely solves the problem, but there are probably better ways of doing it -- I just can't figure out what they are.

### DIFF
--- a/src/discovery.config.js
+++ b/src/discovery.config.js
@@ -16,6 +16,7 @@ require.config({
   }()),
 
 
+  //this will be overridden in the compiled file
   waitSeconds: 30,
 
   // Configuration we want to make available to modules of ths application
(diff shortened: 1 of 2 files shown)


romanchyla:
one of the commenters was saying it doesn't work (multiple requirejs?)

another way to do it, would be inside the main.js before the application gets loaded, here is a hint:

https://github.com/adsabs/bumblebee/blob/master/src/js/apps/bumblebox/bootstrap.js#L109

aholachek:
It does work (I tested it after running grunt release) , but your way is
better I think.

aholachek:
@romanchyla I don't think your solution works for the compiled file after all, since it's a single file (atm, I'm working on splitting it up)

PR #911, opened 2016-05-06
Simbad search

ehenneken:
Implementation of object search, like e,g,
```
bibstem:"A&A" object:((LMC OR SMC) AND Andromeda) year:2015
```
(which should return results)

### DIFF
--- a/src/js/apps/discovery/navigator.js
+++ b/src/js/apps/discovery/navigator.js
@@ -325,7 +325,12 @@ define([
             searchPageAlwaysVisible);
           //allowing widgets to override appstorage query (so far only used for orcid redirect)
           var q = app.getObject('AppStorage').getCurrentQuery();
-          var route = '#search/' + queryUpdater.clean(q).url();
+          if (q && q.get("__original_url")) {
+            var route = '#search/' + q.get("__original_url");
+            q.unset('__original_url');
+          } else {
+            var route = '#search/' + queryUpdater.clean(q).url();
+          }
 
           //taking care of inserting bigquery key here, not sure if right place
           //clean(q) above got rid of qid key, reinsert it
--- a/src/js/widgets/search_bar/search_bar_widget.js
+++ b/src/js/widgets/search_bar/search_bar_widget.js
@@ -295,7 +295,6 @@ define([
           "keypress #search-form-container" : function(){
             //for analytics
             //this is not exact (might have experimented then closed the builder)
-            // but it's close enough
             this._queryBuilderUsed = true;
           }
         },
@@ -326,7 +325,6 @@ define([
             this.$(".q").val(v);
           }
           this.toggleClear();
-
         },
 
         setNumFound : function(numFound){
@@ -410,7 +408,6 @@ define([
               newVal = operator + "(" + currentVal + ")";
               currentVal = "";
               this.setFormVal(newVal);
-              this.toggleClear();
               return
             }
 
@@ -441,9 +438,6 @@ define([
               this.$input.selectRange( newString.length );
             }
           }
-
-          //figure out if clear button needs to be there
-          this.toggleClear();
 
           analytics('send', 'event', 'interaction', 'field-insert-button-pressed', df);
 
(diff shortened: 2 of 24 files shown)


aholachek:
hi @ehenneken, so far this pull request translates from object:obj1 --> simbid:123.
But can it translate the reverse way around? e.g. simbid:123 --> object:obj1

Right now, it seems like

ehenneken:
Hi Alex

When you submit a query simbid:1234 the actual object search will get skipped because it only looks for "object:" in the query string, and in addition "simbid:1234" is a valid query because "simbid" is a valid Solr field. But maybe I'm misinterpreting your question. The micro service as such can translate in both directions.

I'll be happy to explain in more detail tomorrow.

Edwin

aholachek:
Ah ok, I just checked out your branch and if you make a search, then reload the page, you no longer see object:foo, you see simbid:123. Is that ok, or was the intention to hide this from the user?

adsabs-beer-user:
Strange! The original query should remain in the query box. I'll look at it tomorrow!

ehenneken:
Hi @aholachek 

I thought if I do the following to store the original query (e.g. "object:Andromeda") like
```
https://github.com/ehenneken/bumblebee/blob/f5a9cfda326d5271d484230a53fb2b4acffa07b8/src/js/widgets/search_bar/search_bar_widget.js#L489```
and then use it like
```
https://github.com/ehenneken/bumblebee/blob/f5a9cfda326d5271d484230a53fb2b4acffa07b8/src/js/widgets/search_bar/search_bar_widget.js#L322
```
the query box would display that original query "object:Andromeda" (rather than what it was replaced by, i.e. "simbid:1575544")

**Comment by ehenneken** [comment]
The SIMBAD search machinery has been moved to the `query_mediator` and the URL keeps the original `object:....` search in it. But with a hard reload the search box still gets `simbid:...` in it. Working on that...

**Comment by aholachek** [comment]
Hi @ehenneken, there is some unexpected behavior from the object api parser. When I search "simbid0(LMC OR SMC) planet" the api endpoint returns the string "simbid0(LMC OR SMC) planet" which bumblebee cannot handle. I'm not sure why the : gets replaced with a 0.

Would you mind squashing all object search commits into 1 and rebasing on top of the current master?

**Comment by ehenneken** [comment]
the query should be "object:'LMC OR SMC'" rather than "simbid:'LMC OR SMC'"

I'll have a look, probably not before Thursday, and squash all the commits.

**Comment by ehenneken** [comment]
Hi @aholachek a query `object:"SMC OR LMC" bibstem:"ApJ"` works as expected, and a query `object:"SMC OR LMC" planet` also in the sense that it returns 0 results (unlikely you'll find something about a planet in papers about the Small and Large Magellanic Clouds (galaxies). I'll let you know when I've squashed the commits and rebased with master.

**Comment by aholachek** [comment]
Oops I copy pasted the wrong thing, the query object:(LMC OR SMC) planet is throwing an error of "INVALID_SYNTAX: Syntax Error: Unknown function "simbid0" please use this tool to fix it." in your current branch

File: src/js/apps/discovery/navigator.js [hunk 1]
```diff
@@ -325,7 +325,12 @@ define([
             searchPageAlwaysVisible);
           //allowing widgets to override appstorage query (so far only used for orcid redirect)
           var q = app.getObject('AppStorage').getCurrentQuery();
-          var route = '#search/' + queryUpdater.clean(q).url();
+          if (q && q.get("__original_url")) {
+            var route = '#search/' + q.get("__original_url");
```
aholachek: Would there be a way to call queryUpdater.clean() on the __original_url before you store it? Otherwise you could potentially put a very long url in the address bar, full of parameters that are supposed to stay hidden. (E.g. if you search for object:LMC then apply a filter using the author network, the url will be full of __bigquery parameters)

src/js/widgets/search_bar/search_bar_widget.js
```diff
@@ -316,7 +320,11 @@ define([
         },
 
         setFormVal: function(v) {
```
aholachek: Since several other functions depend on using setFormVal, I think this change to the function will prevent other things in the search widget (quick add buttons, query mediator) from working correctly (but I could be wrong)

test/mocha/js/widgets/search_bar_widget.spec.js
```diff
@@ -398,6 +402,21 @@ define([
 
     });
 
+    it("check if the SIMBAD 'object:' search is properly translated into 'simbid:' search", function() {
+      var widget = _widget();
```
aholachek: since most of the translation logic was moved to the query_mediator, could you add tests in query_mediator.spec.js?
I'm possibly (probably) confused but it seems like this particular test isn't testing anything object-query specific, just showing that a start_search event is emitted when the submit button is clicked

PR #1021, opened 2016-07-15
Run saucelabs via Travis CI

jonnybazookatone:
So these changes should allow Travis CI to trigger Saucelabs to use its browser fleet to run our tests that have been defined in the `Gruntfile.js`. Currently, they will only be triggered on a merge into master, or one of its branches. This is because if someone upstream makes a PR, they can potentially do malicious things with the encrypted API keys.

If it's just a PR, then normal `grunt test` should be carried out by travis, otherwise `grunt sauce` is executed.


jonnybazookatone:
Just fyi, I tested this on my local branch and on the saucelab accounts, and it was working.

jonnybazookatone:
Should be fine to merge and see if it works correctly....

PR #1027, opened 2016-07-18
a new, friendlier error message

aholachek:
I think it might be less frustrating for users if we don't confuse them with the details of which part of the api request specifically failed.

Here is the proposed new message:

"There’s something wrong with our backend services.
For the moment, please try <a href="http://adsabs.harvard.edu/abstract_service.html"> the classic service</a> instead.
As we work on scaling up our new generation backend, we occasionally have issues with machines getting overwhelmed, although this will become less of a problem as time goes on."

### DIFF
--- a/src/js/wraps/discovery_mediator.js
+++ b/src/js/wraps/discovery_mediator.js
@@ -21,6 +21,10 @@ define([
     analytics
     ) {
 
+    var genericErrorMessage = 'Our server returned an error.\n' +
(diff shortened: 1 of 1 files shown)


adsabs-beer-user:
Some modified words to consider for the error message:

The ADS server generated an error.  Typically, this is caused by a high
load and it be resolved as we scale up our new backend.  For the moment,
please try the classic service
<http://adsabs.harvard.edu/abstract_service.html> instead.


Steve

aholachek:
Thanks Steve, I edited the message based on your input

Issue #1003, opened 2016-07-12
wrong highlight query


romanchyla:
when searching for `*:* -data:*`

bbb constructs wrong query; `q` will be correct, but `hl.q` is not

```
fl:title,abstract,bibcode,author,keyword,id,citation_count,[citations],pub,aff,volume,pubdate,doi,pub_raw,links_data,property,email
hl:true
hl.fl:title,abstract,body,ack
hl.maxAnalyzedChars:150000
hl.q:-
hl.requireFieldMatch:true
hl.usePhraseHighlighter:true
q:*:* -data:*
rows:25
sort:date desc, bibcode desc
start:0
```

aaccomazzi:
Also failing are queries involving a wildcard and other boolean operators for the same reason, e.g.
`orcid_pub:* OR author:accomazzi`
Which results in:
```
fl:title,abstract,bibcode,author,keyword,id,citation_count,[citations],pub,aff,volume,pubdate,doi,pub_raw,links_data,property,email
hl:true
hl.fl:title,abstract,body,ack
hl.maxAnalyzedChars:150000
hl.q:OR author:accomazzi
hl.requireFieldMatch:true
hl.usePhraseHighlighter:true
q:orcid_pub:* OR author:accomazzi
rows:25
sort:date desc, bibcode desc
start:0
```

PR #1032, opened 2016-07-21
remove hl.q when there is an unbounded wildcard

aholachek:
#1003


aholachek:
@romanchyla before I merge this can you take a quick look and let me know if this solution is ok for now?

Issue #1061, opened 2016-08-09
metrics should by default request the "simple" variant


aholachek:
with a new limit of 8000

Issue #495, opened 2015-07-17
metrics is affected by the solr max rows


romanchyla:
We want to switch to sane conservative limits for /query - the rows parameter.

The metrics widget, however, right now is collecting the bibcodes in one swoop (e.g. rows=300, up to maximum 2000) and then sends them to metrics, possible solutions:

1. display a message "if  you want to get more, you need to login" (but this would require changes in the permission system, so it's more for the future)
2. collect the bibcodes using paging
3. send metrics the query and let it collect the bibcodes

The (2) is more work, but possibly cleaner; (3) is dirt cheap but I'd have to submit patch (metrics service right now uses only 'q' parameter)

romanchyla:
the only advantage to (2) would be that we can limit the amount of bibcodes sent from the client, but i realized I could do the same with (3) if I make metrics accept rows limits and the full query

vsudilov:
Just keep in mind that metrics is allowed to access rows=very many when it queries solr; we don't want to create a backdoor where a user can take advantage of that

Issue #685, opened 2015-11-21
Adjust metrics for single paper to remove unneeded categories


aaccomazzi:
This is a minor issue, but still...
Right now we show a color-coded 4-way breakdown for the citations:
1. refereed to refereed
2. non-refereed to refereed
3. refereed to non-refereed
4. non-refereed to non-refereed

In the case where we look at metrics for a single paper, the possibilities are reduced to two of such categories, either 1+2 or 3+4 (same goes for the reads).  Can we remove the unnecessary legend from the plots (but still keep the colors consistent with our 4-category scheme).  I think this might reduce confusion among other things.

PR #1079, opened 2016-08-24
only requesting tori/riq/indexes if > 50000 citations if user explicitly requests it

aholachek:
addresses #1061 and #495


aholachek:
also addresses  #685

PR #1080, opened 2016-08-24
Citationhelper

ehenneken:


src/js/widgets/citation_helper/widget.js
```diff
@@ -0,0 +1,260 @@
...
+    },
+
+    processSuggestions: function (response) {
+      //how is the json response formed? need to figure out why attributes is there
+      response = response.attributes ? response.attributes : response;
+      if ((response.Error && response.Error === "Unable to get results!") || (response.status == 500)) {
+        this.model.set('msg', 'Citation Helper did not find any suggestions.')
+      } else {
+        this.model.set('items', response);
+        this.model.set('libraryID', this.view.model.libraryID);
```
aholachek: I'm not sure what these lines are doing. On line 82, you can see that this.model == this.view.model

aholachek: in addition, unless I'm really confused, this.view.model.permission will always be undefined; you need to use the get method:
this.model.get("libname")

src/js/widgets/library_individual/widget.js
```diff
@@ -215,6 +215,13 @@ define([
                         data.widgetName = "Metrics";
                         data.additional = {};
                         break;
+                      case "citation_helper":
+                        data.widgetName = "CitationHelper";
+                        data.additional = { libid: id,
+                          permission: this.headerModel.attributes.permission,
```
aholachek: should be this.headerModel.get("permission")

src/js/widgets/citation_helper/widget.js
```diff
@@ -0,0 +1,260 @@
...
+
+    libraryAdd : function(){
+
+      //show loading view
+      this.$(".submit-add-to-library").html('<i class="fa fa-spinner fa-pulse"></i>');
+
+      var data = {};
+
+      //we have the selected library in the model but only if there was a select event, so query DOM
+      data.libraryID = this.$(".submit-add-to-library").val();
```
aholachek: It seems that this value will already be in the model, (you would get it with this.model.get("libraryId") so there is no need to put it into the DOM and then retrieve it.
In general, with backbone you want to limit querying the DOM for data and keeping state in models/collections instead

src/js/widgets/citation_helper/widget.js
```diff
@@ -0,0 +1,260 @@
...
+    viewEvents: {
+      'close-widget': 'closeWidget',
+      "library-add" : "libraryAddSubmit"
+    },
+
+    initialize: function (options) {
+
+      options = options || {};
+
+      this.model = new Model({widgetName : "CitationHelper", libraryID: null, "permission": null, "libname": null});
```
aholachek: You can set all these values as defaults on the model above when you define its class.
It's nice to specify all defaults there because it serves as a sort of documentation for the values that will be put into the model later.

src/js/widgets/citation_helper/widget.js
```diff
@@ -0,0 +1,260 @@
...
+    processResponse: function (response) {
+      //We received bibcodes from the search end point
+      if (response instanceof ApiResponse){
+        var bibcodes = _.map(response.get("response.docs"), function(d){return d.bibcode})
+        // let container view know how many bibcodes we have
+        this.view.model.set({"numFound": parseInt(response.get("response.numFound")),
+          "rows":  parseInt(response.get("responseHeader.params.rows"))});
+        //disable option to get more/fewer bibcodes if there's only 1
+        if (parseInt(response.get("response.numFound")) === 1){
+          this.containerModel.set("requestRowsAllowed", false);
```
aholachek: I think this might be a copy paste remnant-- I see no other reference to containerModel in the widget.
Also, "requestRowsAllowed" is not used by this widget, correct?

src/js/widgets/citation_helper/widget.js
```diff
@@ -0,0 +1,260 @@
...
+        options : {
+          type : "POST",
+          contentType : "application/json"
+        }
+      });
+
+      //normally this info would come from apiquery
+      // let container view know how many bibcodes we have
+      this.view.model.set({
+        numFound : bibcodes.length,
```
aholachek: it seems this information is not currently being used by the widget -- is there a plan to use it later?

src/js/widgets/citation_helper/widget.js
```diff
@@ -0,0 +1,260 @@
...
+        rows : bibcodes.length
+      });
+
+      this.model.set("requestRowsAllowed", false);
+
+      this.view.model.libraryID = data.libid;
+      this.view.model.permission = data.permission;
+      this.view.model.libname = data.name;
+
```
aholachek: the previous lines will fail, you need to do this.model.set({ x : foo, y : boo})

you can check out the nicely annotated source if you want to see how the model is actually storing the data:http://backbonejs.org/docs/backbone.html

src/js/widgets/citation_helper/widget.js
```diff
@@ -0,0 +1,260 @@
...
+      options.library = data.libraryID;
+      options.bibcodes = data.recordsToAdd;
+      name = data.libname;
+
+      //this returns a promise
+      this.getBeeHive().getObject("LibraryController").addBibcodesToLib(options)
+          .done(function(response, status){
+            var numAlreadyInLib = response.numBibcodesRequested - parseInt(response.number_added);
+            that.model.set("feedback", {
```
aholachek: this feedback will never be visible to the user because it is not rendered in the template

aholachek:
screen shot 2016-08-24 at 5 21 30 pm

#### Just a few quick thoughts to keep this widget consistent with the styling of BBB in general:

- I would bold the titles and possibly increase their font sizes since that is information that the user will be looking for first.
- Put a margin bottom on each entry to split it up and make it easier to scan
- Bring the input box a little closer to the entry title to make the association between the two more obvious
- I might add a little padding to the widget, right now all the content is floating a little too close to the edges
- I'd also pay attention to alignment: you want a hard left-hand line where everything lines up. Right now it is jagged, with the input boxes indented a bit.

These points may seem comparatively minor but I think they're still important, since one aspect of usability is good basic visual design.

src/js/widgets/citation_helper/widget.js
ehenneken: That's correct on both counts :-)

src/js/widgets/citation_helper/widget.js
ehenneken: I checked in the console in Chrome and I thought that this.view.model.permission had value "Owner"

src/js/widgets/citation_helper/widget.js
ehenneken: Nope, another copy remnant. It can go.

src/js/widgets/citation_helper/widget.js
ehenneken: I guess it can be just removed, since no feedback needs to be sent to the user (I think)

aholachek:
Oh I forgot to say -- good job in general getting everything to work, it's a relief that bumblebee is at least not totally opaque for someone who's just starting to work with the code. I'll be accessible online this week so maybe after you address these comments I can have another look?

aholachek:
You also need to add the line 'test/mocha/js/widgets/citation_helper_widget.spec.js' to ui_suite.js to get the automated tests to run

src/js/widgets/citation_helper/widget.js
aholachek: I actually would think that showing a brief "success -- 3 papers added to your library!" or an error message if the promise is rejected might be a nice touch, right now it's not totally clear to the user if anything has changed, unless they know to look at the number of records in the upper left hand corner.

src/js/widgets/citation_helper/widget.js
ehenneken: ah, yes, that's what I did! I'll implement the `get` and `set`.

src/js/widgets/citation_helper/widget.js
ehenneken: Exactly, I realized that too. It will make things less mysterious :-)

src/js/widgets/citation_helper/widget.js
```diff
@@ -0,0 +1,244 @@
...
+      var request =  new ApiRequest({
+        target: ApiTargets.SERVICE_CITATION_HELPER,
+        query: new ApiQuery({"bibcodes" : bibcodes}),
+        options : {
+          type : "POST",
+          contentType : "application/json"
+        }
+      });
+
+      this.model.set({'requestRowsAllowed': false,
```
aholachek: I think you can get rid of the requestRowsAllowed parameters since they dont seem to apply to this widget

aholachek:
Hi Edwin,

Looks good! I think you have 2 tests failing now and you need to squash the two commits.
I also think you probably need to hide the citation helper option for the public library view.

ehenneken:
Thanks Alex!

I'll fix the last issues today. I'm hiding the "Add to library" button for non-owner views of a library, but I think having the Citation Helper as such available for public views will be fine.

aholachek:
The button is not hidden for me when I view a private library

ehenneken:
I'm guessing that the "Delete" buttons won't appear in the public view of a library. I'll have to implement the same test for the "Add" button in the Citation Helper.

ehenneken:
The "permission" part for the "Add" button works as it should, I think. When I'm logged into my account and go to a "public library" like `https://ui.adsabs.harvard.edu/#/public-libraries/ScnjZhZDR7Okl8YDRUAhnA`, the "Add" button does not appear after running the Citation Helper

Issue #1121, opened 2016-11-30
Update doc and diagnostics


spacemansteve:
README is slightly out of date for OS X setup.
Diagnostics should not assume core name of collection1.
Why not expose bbb object as a global based on debug parameter in url rather than the more obscure DynamicConfig?

PR #1122, opened 2016-11-30
#1121 improved doc and diagnostics

spacemansteve:
updated README to reflect current versions, etc.
in diagnostics, removed hard coded collection name
exposed bbb as a global object based on debug=true in url, I never did figure out where or how to modify the dynamic configuration.

### DIFF
--- a/src/js/apps/bumblebox/main.js
+++ b/src/js/apps/bumblebox/main.js
@@ -78,11 +78,10 @@ define([
               app.start(Router);
               pubsub.publish(pubsub.getCurrentPubSubKey(), pubsub.APP_STARTED);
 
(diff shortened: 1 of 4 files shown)


src/js/apps/bumblebox/main.js
```diff
@@ -78,11 +78,10 @@ define([
               app.start(Router);
               pubsub.publish(pubsub.getCurrentPubSubKey(), pubsub.APP_STARTED);
 
-              var dynConf = app.getObject('DynamicConfig');
```
romanchyla: so this change is somewhat controversial; exposing the global object is akin to making it available to other scripts to play with - which should be not allowed in production (once we run in production mode)

Issue #1134, opened 2017-01-06
Citations Metrics - Citation Histogram


ehenneken:
When you switch the citation histogram to normalized citations, the numbers in the mouse-overs are still those for the regular citations.

PR #1171, opened 2017-05-22
Histogram Chart Now Shows Correct Data When Normalized

thostetler:
Fixes: #1134


thostetler:
This truncates each data point to 2 decimal places (there was no rounding/truncating previously).

PR #1175, opened 2017-05-30
Bugfix: Close Persistent Error Message After Navigation

thostetler:
* Error message(s) persisting after navigating from the login page now close correctly
* Implemented `closeView` method


thostetler:
Fixes #838

thostetler:
Tests are passing locally, restarting...

Issue #1178, opened 2017-06-02
Application Not Redirecting After Login


thostetler:
After attempting to login, sometimes the page does not redirect to the previous/landing page.
It seems to hang.

PR #1179, opened 2017-06-02
Bugfix: Redirect to Landing Page If No Previous Page

thostetler:
* If the history array was empty, the redirect would be undefined.
* This defaults to the landing page in such a case

#1178


thostetler:
Closing until I can figure out why tests are failing on Travis but not locally

Issue #1162, opened 2017-04-20
Make abstract pages preview-friendly


aaccomazzi:
Applications such as Twitter, Facebook and Slack use metadata embedded in web pages to generate a preview of the page itself.  We were reminded of this by one of our fans at LSST, with a link to this how-to page: https://medium.com/slack-developer-blog/everything-you-ever-wanted-to-know-about-unfurling-but-were-afraid-to-ask-or-how-to-make-your-e64b4bb9254 

See also #792 for additional use cases of embedded metadata.

timj:
Fantastic. Do you know when this code is planned for release?

ehenneken:
It will be available on our development UI (devui.adsabs.harvard.edu) soon, later today. We don't have a date for releasing on production yet.

timj:
I'm happy to try it from Slack when it's deployed to devui (it doesn't seem to work at the moment).

aaccomazzi:
Still an issue at least for slack, see e.g. https://devui.adsabs.harvard.edu/#abs/2018NIMPA.896...24B/abstract

thostetler:
yeah because the crawlers for slack, facebook, twitter, etc. only download html, our dynamic metatags don't get loaded.  We will have to pre-render or generate this metadata server-side in order for unfurl links to work properly.

gpdf:
Still very interested in this.  It's really an obstacle to communicating to people by posting `adsabs` links.

aaccomazzi:
So we are making progress on this under the hood, and we plan to release a service that creates HTML pages and caches them so that they will be crawlable and properly unfurl.  We have already released the building blocks required for this: a push-state version of the UI (which makes abstract pages URL look like this: https://ui.adsabs.harvard.edu/abs/2018NIMPA.896...24B/abstract) and a rendering/caching layer which we are currently testing (https://github.com/adsabs/turbobee).

Unfortunately the work has been delayed due to scheduling issues, but we hope to release this functionality before summer.

Feel free to leave more details on the relevant repos if you have additional thoughts.

timj:
Am I right in thinking that this now works properly? I do see unfurling.

aaccomazzi:
Yes, this was implemented and deployed back in September.  Thanks for the reminder and sorry for being sloppy about closing issues...

PR #1190, opened 2017-06-20
Add Additional Meta Tags For Twitter Cards/Open Graph

thostetler:
* This should *hopefully* allow services like twitter, facebook, slack, etc. to properly unfurl links on the abstract pages
* Unable to test this properly until this is merged due to only having the changes locally.

- #1162


thostetler:
So after getting a chance to test this out, I don't think it's possible to do this with our current system.  Since Twitter and Slack crawlers don't run javascript, and our tags are dynamically generated...

Issue #1210, opened 2017-07-25
Do not send queries with empty quotes


taylorshaulis:
If I execute the query

> abs:""

Bumblebee generates an error Syntax Error, cannot parse abs:""  Try looking at the search examples on the home page or reading our help page.

This error is generated after the search has been sent to the api and the api returns error.

I propose bumblebee parse the query before sending to api and if any empty quotes are present, generate error to user and DO NOT send to API.

Just in the past 10 minutes I have seen abs:"" and bibstem:"" so I believe its worth it to block any ""

PR #1213, opened 2017-07-27
Add api query validator

thostetler:
Added a new component to run some common validations on the search bar input.  Now queries that contain empty `""` or `"^"` or `("")` will be caught on the front-end, instead of relying on the api to validate.  However, more complex queries such as nested fields and garbled input will be passed along to the api to parse.  When the validator catches some bad input, an error message is provided to the user.

- Fixes: #1210

### DIFF
--- a/src/js/components/query_validator.js
+++ b/src/js/components/query_validator.js
@@ -1,0 +1,139 @@
+define([
+'underscore'
+], function (_) {
+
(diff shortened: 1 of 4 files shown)


taylorshaulis:
YAY!!!!!!!!!!!!!!

Issue #1239, opened 2017-08-22
Highlights Broken


thostetler:
Regression after fixing the problem with markup, the highlighted areas of the highlights are no longer showing up.

PR #1240, opened 2017-08-22
Revert Highlight HTML Sanitization

thostetler:
* Don't sanitize highlights html, since this strips out the <em></em> tags that are provided by the api

Fixes: #1239


romanchyla:
a perfect candidate for a unittest ;)

PR #1244, opened 2017-08-23
Fix metrics redirect issue

thostetler:
* Added `componentParams` property to widgets when they are added to the dom
  * Now you can add params using `<div data-widget="showMetrics" data-test="test">`
* Added some checks to make sure not to redirect if on abstract page

The metrics widget should probably be refactored to work better as a standalone widget AND a background widget that is loaded on the abstract page.  Currently it loads everything and just hides graphs it doesn't need.

### DIFF
--- a/src/js/widgets/metrics/widget.js
+++ b/src/js/widgets/metrics/widget.js
@@ -698,11 +698,11 @@ define([
     signalShowAll: function () {
       this.$(".show-all").html('<i class="icon-loading"/>&nbsp;&nbsp;' + this.$(".show-all").html());
       this.trigger('show-all');
-      },
+    },
 
-      modelEvents : {
-        change : 'render'
-      },
+    modelEvents : {
+      change : 'render'
+    },
 
     regions: {
       papersGraph: "#papers .metrics-graph",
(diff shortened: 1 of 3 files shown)


src/js/widgets/metrics/widget.js
```diff
@@ -761,7 +761,11 @@ define([
 
     closeWidget: function () {
       this.reset();
-      this.getPubSub().publish(this.getPubSub().NAVIGATE, "results-page");
+
+      // only redirect if we are NOT on the abstract page
+      if (this.componentParams && !this.componentParams.abstract) {
```
romanchyla: so who is setting .abstract?

thostetler: In the template https://github.com/adsabs/bumblebee/blob/14cc2df3360c13f00cd654800043c97f00495620/src/js/wraps/abstract_page_manager/abstract-page-layout.html#L36

could probably find a better way, wasn't sure if there would be a reason to pass the param to another widget that is on that page.

romanchyla: i see, I'd suggest to make it slightly more abstract (and let other widgets have the same)

```<div data-widget="ShowMetrics" data-allow-redirect="true"/> ```

but if we are in it, you may want to explore WHETHER the navigator should handle those situations

PR #1247, opened 2017-08-25
Vagrant upgrade to latest Ubuntu LTS

marblestation:
Puppet was also removed to simplify vagrant setup


taylorshaulis:
This is way more streamline and much more simple. I love it!

PR #1248, opened 2017-08-28
Fix classic form boolean logic

### LINKED ISSUE #1212: Classic form: enable "simple logic" options
What ADS classic calls "simple logic" is the use of "+" and "-" to require or exclude a particular word in the result list.  If I now use this on a search in the author field such as `+kurtz, m; -accomazzi, a` (on separate lines) the form creates the query `author:("+kurtz, m" "-accomazzi, a")` in which the "+" and "-" signs are ignored by the search engine.  The query should instead by `author:(+"kurtz, m" -"accomazzi, a")`

thostetler:
fixes #1212

### DIFF
--- a/src/js/widgets/classic_form/widget.js
+++ b/src/js/widgets/classic_form/widget.js
@@ -123,9 +123,43 @@ define([
         if (val !== ""){
           logic = $t.find(".logic-group input:checked").val();
           field = $t.data("field");
-          if (logic == "BOOLEAN"){
-            val = val.replace("\n", " ");
-            qDict.q.push(field + ":(" + val + ")");
+          if (logic === "BOOLEAN"){
+
+            // create a new updater object
+            var updater = new ApiQueryUpdater(' ');
+
+            // expecting input to be on separate lines, with each proceeded by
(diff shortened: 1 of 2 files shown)


src/js/widgets/classic_form/widget.js
```diff
@@ -123,9 +123,41 @@ define([
...
+
+              // test the first character for presence of (+ or -)
+              if (/[+\-]/.test(line[0])) {
+
+                // +blah -> +"blah"
+                line = line[0] + '"' + line.slice(1) + '" ';
+              } else {
+
+                // blah -> +"blah"
+                line = '+"' + line + '" ';
```
romanchyla: please use the following:

```updater.quoteIfNecessary(line)```

https://github.com/adsabs/bumblebee/blob/0b242c9ecaf1e831c5673faf0f72b72b0ed6957b/src/js/components/api_query_updater.js#L270

the quotes do change how the search engine handles stuff (in some situations); so better to play safe and not use quotes if we don't have to

thostetler:
Updated the PR.

PR #1401, opened 2018-03-09
Update Url Parsing

thostetler:
- Add the ASCII encoding to the characters the hash string is split on


romanchyla:
@thostetler i'm really unsure this change is good; cause the url can contain ? and also a redirect query that  has the encoded ? - IMO nothing should modify the other, double encoded portion. but i might be misunderstanding it - what is this change trying to address?

PR #1435, opened 2018-04-12
Added error recovery

romanchyla:
this is not ready yet; mocha is failing (something stupid - not even related to a test)

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ dev setup - linux
 =================
 
 ```bash
-  $ sudo apt-get install node npm phantomjs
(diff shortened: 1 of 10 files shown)


romanchyla:
Ready for review; I had to deactivate one test - after having spent close to 4 hours trying to isolate source of a problem - it seems like the new code, which added automated retry - registers a new ajax call, which holds the query mediator from being recycled (when we call qm.destroy() inside the test)

romanchyla:
gosh, it was all passing in the browser - then i rerun `grunt test`; 64 failures - then 3 failures; and browser fails 3; sigh...distance is needed

PR #1449, opened 2018-04-24
Add Bootstrap To Local Storage

thostetler:
Store bootstrap config into localStorage, and use it instead of new
if it hasn't expired.

In my tests this speeds up subsequent loads (while token is not expired) by 300-400ms

### DIFF
--- a/src/js/mixins/discovery_bootstrap.js
+++ b/src/js/mixins/discovery_bootstrap.js
@@ -81,8 +81,14 @@ define([
       var beehive = this.getBeeHive();
       var dynConf = this.getObject('DynamicConfig');
       var timeout = 3000;
-
       var defer = $.Deferred();
+
+      // check out the local storage to see if we have a copy
+      var storage = beehive.getService('PersistentStorage');
+      var config = storage.get('appConfig');
+      if (config && config.expire_in && new Date(config.expire_in) > new Date()) {
+        return defer.resolve(config).promise();
+      }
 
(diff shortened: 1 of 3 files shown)


thostetler:
Don't merge until this is reviewed by @romanchyla, thanks!

src/js/mixins/discovery_bootstrap.js
```diff
@@ -81,9 +81,15 @@ define([
       var beehive = this.getBeeHive();
       var dynConf = this.getObject('DynamicConfig');
       var timeout = 3000;
-
       var defer = $.Deferred();
 
+      // check out the local storage to see if we have a copy
+      var storage = beehive.getService('PersistentStorage');
+      var config = storage.get('appConfig');
+      if (config && config.expire_in && new Date(config.expire_in) > new Date()) {
```
romanchyla: seems good to me @thostetler , it might be advisable to remove the stored token here (provided the upstream puts it back into storage) - or invalidate it elsewhere when we detect the access was denied - otherwise we may end up in a loop always returning the same token (also, the user's machine may have a wrong date; so it can think the expired token is not expired)

another concerns, but that's just for completeness, is to check the permanent storage is cleared when user logs out

thostetler: ah good idea, I'll make a ticket to address those issues.  Thanks for checking it out! 👍

PR #1463, opened 2018-05-04
Update resource links

thostetler:
Rewrote parts of the source links widget.  Now follows a better messaging pattern for redux, gets rid of thunk and immutable.

Small redesign of the widget styling


test/mocha/js/widgets/resources_widget.spec.js
```diff
@@ -241,75 +259,21 @@ define([
...
+            shortName: 'My Institution'
+          }],
           dataProducts: []
         });
         this.pubsub.publish(this.pubsub.DELIVERING_RESPONSE, mockResponse);
         const icon = $('i.fa-university', $el);
         expect(icon.length).to.eql(1);
         done();
       });
-      it('Shows a "Show All" button if there are more than 3 of either kind of source', function (done) {
```
romanchyla: is this functionality gone? done differently?

src/js/widgets/resources/components/app.jsx.js
```diff
@@ -0,0 +1,133 @@
+define([
```
romanchyla: i have only a generic comments; they can't really bear too much weight on the review as I dind't invest any time lerning react

so, having admitted of being an ignorant, i'll proceed on to kibicking ;-)

if this as a react way of buiding views (and I'd like to point out it mixes presentation; like classnames), then it's ok - if it is normal in react world to mix logic with presentation, then i'm clearly way off the current trends (i.e. i don't like 'em!)

also, it feels so much more verbose - i.e. it's long - can that be helped?

src/js/widgets/resources/containers/app.js
```diff
@@ -0,0 +1,29 @@
...
+  'es6!../components/app.jsx'
+], function (_, ReactRedux, ui, App) {
+
+  // actions
+  const {
+    handleLinkClick,
+    handleResetClick
+  } = ui;
+
+  // mapping state to props
```
romanchyla: i *think* i like this; more use of stores to communicate changes in general imho is good

src/js/widgets/resources/redux/middleware/api.js
```diff
@@ -0,0 +1,155 @@
...
+    SET_FULL_TEXT_SOURCES,
+    SET_DATA_PRODUCTS,
+    LINK_CLICKED
+  } = ui.actions;
+
+  /**
+   * Fires off request, delegating to the outer context for the actual
+   * fetch
+   */
+  const fetchData = (ctx, { dispatch }) => next => action => {
```
romanchyla: all the parts here; i have a feeling they will be repeated/similar in other widgets - but this middleware seems to be unique to resources

can it be abstracted? also, if you search for 'Loading' strings - those parts seem to be common to any widget too

test/mocha/js/widgets/resources_widget.spec.js
thostetler: yeah, originally I was thinking that there would be papers with tons and tons of data sources.  In practice, however, I don't think there are any articles with more than about 12 or so at the max, so I just list them out now in two columns.

src/js/widgets/resources/components/app.jsx.js
thostetler: I agree with you about mixing logic, but it's almost the same thing as using handlebars or some other templating language.  I tried to not do any reformatting or computing within the render blocks, only doing conditional checks.

src/js/widgets/resources/containers/app.js
thostetler: Yeah the idea is that reactredux will shallow compare the properties we list here and figure out if it needs to render based on a change in state.  It also allows us to restrict what props make it down to the component, so we could add a bunch of containers if we wanted different parts to render independently.

src/js/widgets/resources/redux/middleware/api.js
thostetler: I was thinking about that -- I think it would be cool to write a generic reducer for the connection between the widget and the react component.

PR #1467, opened 2018-05-08
support for tugboat messages

spacemansteve:
when translating a classic query, tugboat may add error and warning messages. these changes support bumblebee display of the mesages.
this is a partial solution, it does not include display of tugboat messages for objects searches.
this commit includes the qs function from utils.js because it is not in this branch.  at some point the qs function from utils should be used.

### DIFF
--- a/src/js/components/query_mediator.js
+++ b/src/js/components/query_mediator.js
@@ -18,8 +18,8 @@ define(['underscore',
     'js/components/api_feedback',
     'js/components/json_response',
     'js/components/api_targets',
-	'js/components/api_query',
-	    'js/components/alerts',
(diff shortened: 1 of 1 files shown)


src/js/components/query_mediator.js
```diff
@@ -738,6 +743,40 @@ define(['underscore',
         }
       },
 
+      getAlerter: function() {
+        return this.getApp().getController(this.alertsController || 'AlertsController');
+      },
+
+      // this function will eventually be available in src/js/utils.js
+      qs: function (key, str) {
```
thostetler: So you should be able to bring this in above, doing something like:

PR #1545, opened 2018-07-17
Update No-Script Content

### LINKED ISSUE #1544: Bad link generated in bumblebee/src/index.html if client has no javascript
### Expected Behavior
If the client has no javascript capability, they should get a message that BBB requires JavaScript, and be sent to Classic (http://adsabs.harvard.edu/abstract_service.html)

### Actual Behavior
If the client has no javascript capability (they have javascript turned off, or they have a plugin that is specifically blocking our application domain (harvard.edu), they get the following message:
          <p>
              Welcome hardcode user. Your client does not speak JavaScript,
          please follow this link to access a non-JavaScript version of this
              website:
          <a href="//adsabs.harvard.edu/abstract_service.html">ADS Classic</a></p>

The problem is the lack of a leading "http:" will default this url to "https:" which doesn't work with adsabs.harvard.edu.  The user will wait several seconds, and then get an "unable to connect" message after they click this link.

Secondary problem: did you mean "hardcore" instead of "hardcode" in the message text?  I would omit the first sentence entirely.

### Steps to Reproduce

Turn off javascript in browser preferences, and load https://ui.adsabs.harvard.edu/

thostetler:
Adds a styled <no-script> message for users without javascript
Fixes the protocol issue on the classic link

Resolves #1544

image

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -18,21 +18,54 @@
     type="application/opensearchdescription+xml"
     href="https://ui.adsabs.harvard.edu/opensearch.xml"
     title="ADS Search" />
+  <noscript>
+    <style>
+      #app-container {
+        display: none;
+      }
+      body {
+        overflow: hidden;
+        position: absolute;
+        width: 100%;
+        height: 100%;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+      }
+      #noscriptmsg {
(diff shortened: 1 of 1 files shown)


seasidesparrow:
Great, much better message, thanks Tim!

PR #1596, opened 2018-09-17
Updates to ORCID to work with new orcid-service table

kelockhart:


src/js/modules/orcid/bio.js
```diff
@@ -0,0 +1,50 @@
...
+
+  var Bio = function Bio(bio) {
+    this._root = bio || {};
+
+    this.get = function (path) {
+      var val = jp.query(this._root, path);
+      if (_.isEmpty(val)) {
+        return null;
+      }
+      if (_.isArray(val) && val.length <= 1) {
```
thostetler: nitpick, but technically here you could just return `val[0]` to accomplish the same

thostetler:
Looks good, I'd say go ahead and do a rebase and merge it! 👍

PR #1665, opened 2018-12-08
Lazy loaded widgets

romanchyla:
Proof of concept, implementation of the lazy loaded components 

Still buggy - navigator will have to be updated (a display gets messed up for the landing page - probably because a widget gets inserted into a page after it was already CSS'ed; not sure....)


romanchyla:
I have finished conversion of almost all navigation endpoints; this is rather large PR - and I didn't yet touch unittests; I have however dealt with two competing paradigms in the bbb codebase; router.js had too much of logic - now it is using navigation for everything; and it is asynchronous everywhere

src/js/apps/discovery/navigator.js
```diff
@@ -451,44 +558,179 @@ function (
...
+        var possibleSearchSubPages = ['Metrics', 'AuthorNetwork', 'PaperNetwork',
+        'ConceptCloud', 'BubbleChart'];
+
+        var widgetName, pages;
+
+        // convention is that a navigate command for search page widget starts with "show-"
+        // waits for the navigate to results page emitted by the discovery_mediator
+        // once the solr search has been received
+        if (data.page)
+          widgetName = _.map(data.page.split('-').slice(1), function (w) { return w[0].toUpperCase() + w.slice(1); }).join('');
```
thostetler: just for fun, I think we can replace this whole thing with something like:

```
widgetName = data.page.replace(/show-([a-z])(.*)/, function (str, a, b) {
  return a.toUpperCase() + b;
});
```

thostetler: arguably better haha

src/js/wraps/discovery_mediator.js
```diff
@@ -423,26 +431,23 @@ function (
...
+            app.getService('Navigator').navigate('results-page', { replace: false })
+              .then(function() {
+                qm.getQueryAndStartSearchCycle.apply(qm, argz);
+              })
+            stupidGoAhead = false;
           }
 
-          qm.getQueryAndStartSearchCycle.apply(qm, arguments);
+          if (stupidGoAhead)
+            qm.getQueryAndStartSearchCycle.apply(qm, arguments);
```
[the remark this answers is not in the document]
romanchyla: i was concerned that i didn't know under what circumstances the first two blocks get called; not sure if moving it would result in the same behaviour - so i resolved to keep it; it's ugly yes

test/mocha/js/apps/discovery/router.spec.js
```diff
@@ -26,110 +28,70 @@ define([
...
-      ]);
-      expect(fakePubSub.publish.args[1][1]).to.eql(
-        "[Alert]-Message"
-       );
-
-      expect(fakePubSub.publish.args[1][2].toJSON()).to.eql({
-        "code": 0,
-        "msg": " <b>fakeError</b> <br/><p>Please try again, or contact <b> adshelp@cfa.harvard.edu for support </b></p>"
+      r.routerNavigate('foo')
+      expect(fakePubSub.publish.args[fakePubSub.publish.args.length-1][1]).to.eql(fakePubSub.NAVIGATE);
```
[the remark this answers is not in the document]
romanchyla: much better, thanks!

src/js/apps/discovery/navigator.js
romanchyla: not my code; so i'm kicking the can down the road...

but notice your proposal would fail for 'show-something-somewhere-elsewhere'

PR #1689, opened 2019-01-18
Update To Work with Push-State Routing

thostetler:
Should use pushstate by default, and also convert all `#/` over when it sees them

Remember to update Discovery.vars.js to set `pushState: true`

### DIFF
--- a/src/js/page_managers/templates/results-page-layout.html
+++ b/src/js/page_managers/templates/results-page-layout.html
@@ -2,7 +2,7 @@
     <div class="row s-stable-search-bar-height">
         <div class="s-search-bar-full-width-container">
             <div class="col-xs-0 col-sm-12 col-md-2 s-back-button-container">
-                <a href="/" class="back-button btn btn-sm btn-default"> <i class="fa fa-arrow-left"></i> Start New Search</a>
+                <a href="#" class="back-button btn btn-sm btn-default"> <i class="fa fa-arrow-left"></i> Start New Search</a>
             </div>
             <div class="col-xs-12 col-sm-9 col-md-7 s-search-bar-row" id="search-bar-row" data-widget="SearchWidget"></div>
         </div>
--- a/src/js/page_managers/templates/results-page-layout.html
+++ b/src/js/page_managers/templates/results-page-layout.html
@@ -2,7 +2,7 @@
     <div class="row s-stable-search-bar-height">
         <div class="s-search-bar-full-width-container">
             <div class="col-xs-0 col-sm-12 col-md-2 s-back-button-container">
-                <a href="#" class="back-button btn btn-sm btn-default"> <i class="fa fa-arrow-left"></i> Start New Search</a>
+                <a href="/" class="back-button btn btn-sm btn-default"> <i class="fa fa-arrow-left"></i> Start New Search</a>
             </div>
             <div class="col-xs-12 col-sm-9 col-md-7 s-search-bar-row" id="search-bar-row" data-widget="SearchWidget"></div>
         </div>
--- a/src/shim.js
+++ b/src/shim.js
@@ -39,22 +39,26 @@
   })();
 
   var setGlobalLinkHandler = function () {
+    var routes = [
+      'classic-form',
+      'paper-form',
+      'index',
+      'search',
+      'execute-query',
+      'abs',
+      'user',
+      'orcid-instructions',
+      'public-libraries'
+    ];
+    var regx = new RegExp('^#(\/?(' + routes.join('|') + ').*\/?)?$', 'i');
+
     // apply a global link handler for push state
     require(['jquery'], function ($) {
       $(document).on('click', 'a', function (ev) {
         if (Backbone.history.options.pushState) {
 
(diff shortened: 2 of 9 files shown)


romanchyla:
when a link contains `#abs/2019LRR....22....1I/abstract` then the function will not be invoked; it gets invoked if the link reads `/#abs/2019LRR....22....1I/abstract`

the regex pattern doesn't match, but also, I *think* backbone will handle only links that starts with the root - in discovery.vars.j we have `/`  (maybe we could experiment with ``)

so either we update our templates to contain links that start with the root '/#abs...' or we change root to ``; and we also need to fix that regex

one additional issue: when i'm on the list of results page, and I click on an abstract -- the function gets invoked multiple times; that is not expected, right? I'd at least think it gets to be called only once `preventDefault` should stop any bubbling of events up

thostetler:
So to answer about using "/" versus "", that doesn't seem to affect anything, I've been playing around with both and it seems to work.

But there is a reason for it matching `#abs...` since we also have to detect any local references, like when a link is actually referring to another element.  

We could look for `/` in the href somewhere, I think that would do it.

romanchyla:
It didn't work for me; for links that didn't start with '/' my browser was opening them in a new tab.

I have slept on it and I think this is waht we should do:

  1. move the function from navigator to router (it belongs there anyways, because it affects the entry/exit points)
  2. make the function sensitive only to paths specified in the router (see below for the list, that matches anywhere in the url)
  3. for matching links, update url; ignore the rest


current routes:

```routes: {
      '': 'index',
      'classic-form': 'classicForm',
      'paper-form': 'paperForm',
      'index/(:query)': 'index',
      'search/(:query)(/)(:widgetName)': 'search',
      'execute-query/(:query)': 'executeQuery',
      'abs/:bibcode(/)(:subView)': 'view',
      /*
        * user endpoints require user to be logged in, either
        * to orcid or to ads
        * */
      'user/orcid*(:subView)': 'orcidPage',
      'user/account(/)(:subView)': 'authenticationPage',
      'user/account/verify/(:subView)/(:token)': 'routeToVerifyPage',
      'user/settings(/)(:subView)(/)': 'settingsPage',
      'user/libraries(/)(:id)(/)(:subView)(/)(:subData)(/)': 'librariesPage',
      'user/home': 'homePage',
      /* end user routes */

      'orcid-instructions': 'orcidInstructions',

      'public-libraries/(:id)(/)': 'publicLibraryPage',
      '*invalidRoute': 'noPageFound'
    },
```

so I'm proposing to have something like:

```
if (href == '') {

else if (href.match(/#(index|search|abs|user)\//)

else if (href.match(/#(classic-form|paper-form|orcid-insturctions)/) { // 
``

**Comment by thostetler** [comment]
Okay, I'll give those changes a shot

**Comment by romanchyla** [comment]
I got the behaviour you described with `#abs/routes` but when I (manually) set those to `/#abs` in the function i was getting the desired behaviour

it might be just order of things - try moving registering the handler to index.html (if router is too far back)

**Comment by thostetler** [comment]
So you think we should use a click handler then? -- I took out that part to work with the router, since we knew the route.

File: src/js/page_managers/templates/results-page-layout.html [hunk 1]
```diff
@@ -2,7 +2,7 @@
     <div class="row s-stable-search-bar-height">
         <div class="s-search-bar-full-width-container">
             <div class="col-xs-0 col-sm-12 col-md-2 s-back-button-container">
-                <a href="#" class="back-button btn btn-sm btn-default"> <i class="fa fa-arrow-left"></i> Start New Search</a>
+                <a href="/" class="back-button btn btn-sm btn-default"> <i class="fa fa-arrow-left"></i> Start New Search</a>
```
romanchyla: I'd prefer to not make any changes to templates; keep non-push-state to be our default; and we can handle it with the click function

src/shim.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+    else {
+      setTimeout(checkLoad, 10);
+    }
+  })();
+
+  var setGlobalLinkHandler = function () {
+    // apply a global link handler for push state
+    require(['jquery'], function ($) {
+      $(document).on('click', 'a', function (ev) {
```
romanchyla: to your question above (about click handler): I had the same as what is here in mind

I only think that we can be more specific - on line 57 we should limit ourselves only to match hash patterns recognized by the router; and leave the rest


btw: I think you want to use event.preventDefault() for when we catch

thostetler: Cool, and yeah as far as I know, when you're in a jquery handler it'll prevent default behavior and propagation by default if you return false.  I figured, we can let the app reload if something fails here, otherwise the link just does nothing, not sure the best behavior.  Could load a 404 instead.

src/js/page_managers/templates/results-page-layout.html
thostetler: Sure I can remove all changes to templates.

src/shim.js
thostetler: Okay I can make that change, maybe it is better to list them explicitly rather than being general here, to make sure we match known routes.

PR #1690, opened 2019-01-22
Adds More Bundles To Build

thostetler:
Includes Landing-Page, Search-Page, and Abstract-Page bundles

During the build step, it will generate a separate configuration for each, then dynamically choose the one to use right after requirejs is loaded


romanchyla:
I made some necessary changes; it now loads

btw - noticed that the page was strangely slow; then i saw then the lazy-loading was reverted (i'm assuming that was something with orcid and will be reverted back) - but it proves that without the lazy loading it is noticeably slower

PR #1710, opened 2019-02-04
Push arXiv ID to ORCID w/ claims

kelockhart:


src/js/modules/orcid/work.js
```diff
@@ -297,6 +298,14 @@ define([
         exIds.values.push(doi);
         exIds.relationships.push('SELF');
       }
+      var arxiv = get(ads.all_ids).find(function(element) {
+        return element.toLowerCase().startsWith('arxiv')});
```
thostetler: do you mind moving the closing curly brace and paren? it makes it a little weird to read

PR #1713, opened 2019-02-07
Solved the loading/hanging issue when opening new views

romanchyla:


thostetler:
I'm not sure about the hanging, you mean it is waiting the second for the handler code to finish before loading the page? 

Also, if we move the call to inside the require like you did, we can remove the timeout, since that was really just because I didn't have access to require in time to load up jquery

romanchyla:
The main problem is that it was trying to load /jquery.js, that's because
the J's mapping was not available to point to the right path. I will remove
the wait, thanks

PR #1716, opened 2019-02-13
Static page updates

thostetler:


src/js/mixins/discovery_bootstrap.js
```diff
@@ -194,9 +194,14 @@ function (
       // get together all pages and insert widgets there
       masterPageManager.assemble(app).done(function() {
         // attach the master page to the body
-        var container = $('div#body-template-container').empty();
-        $(masterPageManager.view.el).hide().appendTo(container).show();
-
+        var $main = $('div#body-template-container');
+        if (window.__PRERENDERED) {
+          var $content = $('div#content-container');
+          $($content.selector, masterPageManager.view.el).html($content.html());
```
romanchyla: i'm not sure about this part, jquery docs say that .html() will replace (remove) both data and events; while previously we were just hiding it

also note, that .show() is not being called any more

this could mess up some interactions inside widgets, not sure

src/js/page_managers/controller.js
```diff
@@ -87,52 +90,38 @@ function ($, _,
...
-            else {
-              // reducing unneccessary rendering
-              if (widget.getEl) {
-                el = widget.getEl();
+            var doRender = function () {
+              // in case the user passed data params on the dom element,
+              // create props on the widget
+              _.assign(widget, { componentParams: $(widgetDom).data() });
+
+              if (window.__PRERENDERED && widget.view && PRIORITY_WIDGETS.indexOf(widgetName) > -1) {
```
romanchyla: i'm going to merge with these changes in, also to try to see the interactions, but could you try to think about something more generic for the future?

i know it is not easy, but if we were to start caching search results or other pages (i.e. /abs/<bibcode>/similar), this can get out of hand fairly quickly -- something maybe as for each widget, on the page, instead of constructing the view, find the appropriate dom, and attach to it (and set that area to be the view) -- like you do with the `handlePrerenderedContent` just tiny bit more generic

PR #1724, opened 2019-02-21
Small push state fixes

thostetler:
Switching `#` links that don't want to redirect to `javascript:void(0)`
Remove `event.preventDefault` and `return false` instead
Other small fixes

### DIFF
--- a/src/js/widgets/classic_form/widget.js
+++ b/src/js/widgets/classic_form/widget.js
@@ -45,7 +45,6 @@ define([
     },
 
     submitForm: function (e) {
-      e.preventDefault();
       var queryDict = this.serializeClassic();
       if (!queryDict.q.length) {
         // allow searching of just dates (which is a filter)
@@ -60,6 +59,7 @@ define([
           $el.html(currHtml);
         }, 3000);
       });
+      return false;
     },
 
     serializeClassic: function () {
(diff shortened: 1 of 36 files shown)


src/js/widgets/classic_form/widget.js
```diff
@@ -60,6 +59,7 @@ define([
           $el.html(currHtml);
         }, 3000);
       });
+      return false;
```
romanchyla: @thostetler - i was also curious about the difference between `return false` and `preventDefault`; according to this SO post the `return false` will cancel the default action for links (navigation), but for forms it might be different; there the default action might be to submit the form

so, i'm not sure what is going to happen here, but I hope you tested it :)

romanchyla: here is the link: https://stackoverflow.com/questions/128923/whats-the-effect-of-adding-return-false-to-a-click-event-listener#answer-128966

thostetler: Yeah, I think he is talking about regular event handlers, in which case returning false doesn't do much other than stop the current handler.  In jQuery though, it is the same as calling `e.preventDefault()` and `e.stopPropagation()`.   https://www.mail-archive.com/jquery-en@googlegroups.com/msg71371.html

I think it should be safe to do this in these handlers, since we aren't concerned with allowing the event to bubble here, and if it gets to that outer handler it may do something unexpected.  

Also, yeah I've been testing out forms and links around and this seems to all work well

PR #1727, opened 2019-02-25
Add in sw-precache service worker

thostetler:
- This is a first pass at service worker caching, more investigation is needed to determine the best configuration.
 - For example, right now we include all the paths, technically we could remove ones not found in bundles, but there is no easy way to determine at build time which bundle will be used.


romanchyla:
There is a set of paths, present in all the bundles - so those could be removed.

However, what is not clear to me is how we are going to run/deploy in production. The grunt hides all the details, my assumption is that it adds endpoints to the webserver; the worker needs to communicate with the server component, no?3

thostetler:
No, it shouldn't add any endpoints.  So the way it works afaik, is that it captures requests from the application (outside of the actual app) and then decides whether to forward them to the server or to give it back whatever is in the cache.

It should basically take every request and reply from the cache first, but also make the request, the app gets an immediate response, but the service worker will wait for the reply and if the response is different than what is in cache it'll invalidate and refresh the page.

romanchyla:
the bit about serving from cache first, then giving updated data is nice - but in the docs they say sw-precache is proactively caching everything

So, say we have landed on the homepage (and loaded landing bundle) -- will the worker be downloading all the assets if we leave the page opened (or even before it shows the page?) -- even those that are part of the bundle? And how about later on, when a new release is made, how does it know that a changed module should be loaded from a (new) bundle or from a network location

i guess i should just try to run it when I have a moment...

thostetler:
It should only download whatever is requested by the app.  The precache is generated at build time by they grunt config.  It looks like there is an option for dynamic pages that will look at the routes and grab a set of paths, I'm not sure how to handle the bundles properly.

romanchyla:
seems to work great! it's magic :)

however, something weird is also going on - the landing page is 800Kb; search page when loaded 1.3MB (just javascript) - compared to un-optimized UI of 1.2

i'm noticing that when on a results/abstract page - some resources are not part of the bundle; the resulting download could perhaps be shrunk little bit if they were included in the bundle (a correct bundle gets loaded) - but i'm not sure that can on itself explain having bigger download (unless bbb got more libraries and is bigger on itself)

do you see similar numbers?

thostetler:
I know! its pretty cool ;) 

Yeah I'm seeing the same sizes on the bundle, I think I know what it is, I altered a couple paths earlier maybe I broke something.  I'll play around with that and see if I can get it back to normal.

PR #1734, opened 2019-03-08
Bugfixes

thostetler:
This was a much larger fix than anticipated.  This includes a number of small improvements to the abstract page nav.  I think in the future this should be refactored to make updates simpler, right now the way the widgets communicate can cause some issues with state.  This PR addresses the issue of stale data persisting, and also it cuts down on some of the superfluous API calls that are made on the abstract page.


src/js/widgets/list_of_things/details_widget.js
```diff
@@ -66,9 +68,12 @@ function (
 
     ingestBroadcastedPayload: _.noop,
 
-    dispatchRequest: function (apiQuery) {
+    dispatchRequest: function (bibcodeOrQuery) {
 
-      this.currentQuery = apiQuery;
+      var apiQuery = bibcodeOrQuery;
+      if (!(bibcodeOrQuery instanceof ApiQuery) && _.isString(bibcodeOrQuery)) {
+        apiQuery = new ApiQuery({ q: 'bibcode:' + bibcodeOrQuery });
```
romanchyla: i'd recommend using `identifier:` index instead; that way even old (and non) bibcodes will work

test/mocha/js/page_managers/toc_manager_test.spec.js
```diff
@@ -111,43 +111,45 @@ define([
                   var references = w['ShowReferences'];
 
                   var r = new ApiResponse(testData());
-                  r.setApiQuery(new ApiQuery({q: 'foo'}));
+                  r.setApiQuery(new ApiQuery({q: 'bibcode:foo'}));
 
                   abstract.processResponse(r);
 
                   app.getWidget("TOCWidget").done(function(widget) {
+                    window.app = app;
```
romanchyla: used for debugging? bbb shouldn't need the global

src/js/widgets/graphics/widget.js
```diff
@@ -94,6 +94,14 @@ define([
         this.loadBibcodeData(bibcode).done(function () {
           self.trigger('page-manager-event', 'widget-ready', { isActive: true, widget: self });
         });
+        this.getResponseDeferred().fail(function () {
+          self.trigger('page-manager-event', 'widget-ready', {
+            closing: bibcode !== this._bibcode,
```
romanchyla: i was confused by the name of this var; given the context it seems to be applied to situations when widget failed to receive data - pls consider something like `isReady` or `isLoaded`

test/mocha/js/page_managers/toc_manager_test.spec.js
thostetler: yep, missed that one I'll take it out

src/js/widgets/list_of_things/details_widget.js
thostetler: @romanchyla The only thing is that there are multiple places that try to parse out `bibcode:.*` so it would break those if we changed to `identifier`.  Unless you think we should refactor those

src/js/widgets/graphics/widget.js
thostetler: yep I agree, although this is just a flag to tell toc_widget to switch to the Abstract widget and to set `isActive` to false.  Could probably be done differently though.  

Also, yeah the only real usecase for this is when the widget needs to close for some reason (i.e. there was an error when loading)

thostetler:
Going to merge for now, I'll make a ticket to refactor `bibcode:` to `identifier:`

PR #1735, opened 2019-03-11
fix issue with search bar disappearing

thostetler:
This is a simple fix, it just moves the swapping out of DOM elements until after all the page manager DOMs are loaded and rendered.


romanchyla:
@thostetler: what happens if one of the widgets takes forever to render? Are we still going to be fine, or is user going to be waiting?

thostetler:
That is a good point, even now though it would make the user wait, since the promises have to all be resolved anyway.  We could wrap them in a timeout, maybe like 10 seconds or something, and retry the render or resolve it.

thostetler:
Merging for now, we can revisit if this causes any issues with timing.  For now, I don't see a big difference between what was there before and this, other than that everything goes all at once now, instead of being swapped as they load.

romanchyla:
I think it will just make the page transitions little bit slower and hard to notice (not a bug per se)

I checked the previous version wasn't synchronous; so it would render the widgets when they were called. Maybe this experience will be less jarring to users, but it might slow down things. I'm going to create an issue for it

Issue #1733, opened 2019-03-07
Extra entry in history for searches


aaccomazzi:
Perform the following query `bibstem:ascl`. The URL changes momentarily to be:
```
https://qa.adsabs.harvard.edu/#search/q=bibstem%3Aascl&sort=date%20desc%2C%20bibcode%20desc&_p=0
```
before being re-written to:
```
https://qa.adsabs.harvard.edu/#search/p_=0&q=bibstem%3Aascl&sort=date%20desc%2C%20bibcode%20desc
```
I now end up having two entries in my history, often (but not always it seems?) leading to a loop: if I hit back I go to the first URL which gets rewritten to the second one, with no end in sight.

Issue #1634, opened 2018-11-20
Firefox history issue when transitioning to abstract page


thostetler:
### Expected Behavior
Should be able to go back to library after clicking on a link
### Actual Behavior
Back button does not seem to work with firefox
### Steps to Reproduce
open any library, click on link, from abstract page use browser back button 
observe that the page does not return to library

Issue #1527, opened 2018-07-02
Back button not returning to correct search form in some browsers


kelockhart:
### Expected Behavior
Go to the Classic Form (https://ui.adsabs.harvard.edu/#classic-form) in either Firefox Quantum (tested using both 60.0.2 or 61.0 on a Mac) or Safari 11.1.1. Fill in the search form and click Search. Click the browser's back button - it should return you to the Classic Form.

### Actual Behavior
Instead, you're returned to the Modern Form. If you click back into the Classic Form, search again, and click the back button again, it will finally take you back to the Classic Form.

### Steps to Reproduce
Able to replicate in Firefox and Safari only (not Chrome, which behaves as expected). Searched for both author and title - the search input doesn't seem to matter.

PR #1745, opened 2019-03-15
More qa fixes

thostetler:
@romanchyla, please check this PR out.  This is based off the other PR, which I'll close.  It removes the check loop in favor of a new pubsub event `UPDATE_CURRENT_QUERY` which it watches for instead.  

I need some guidance on how to keep repeated queries from happening, this line: https://github.com/adsabs/bumblebee/blob/master/src/js/wraps/discovery_mediator.js#L433
I had to remove it, since it was causing a lot of the double-navigation issues.  What would be the best way to fix it?

### DIFF
--- a/src/js/apps/discovery/navigator.js
+++ b/src/js/apps/discovery/navigator.js
@@ -142,6 +142,7 @@ function (
       });
 
       this.set('LibraryImport', function (page, data) {
+        var self = this;
         var defer = $.Deferred();
         if (redirectIfNotSignedIn()) {
           defer.resolve();
@@ -154,6 +155,7 @@ function (
               .done(function (widget) {
                 widget.setActive('LibraryImport');
                 self.route = '#user/settings/libraryimport';
+                self.title = 'Library Import';
                 publishPageChange('settings-page');
                 defer.resolve();
               });
@@ -163,7 +165,7 @@ function (
       });
 
 
-      function settingsPreferencesView(widgetName, defaultView) {
+      function settingsPreferencesView(widgetName, defaultView, title) {
         return function (page, data) {
           var defer = $.Deferred();
 
@@ -185,6 +187,7 @@ function (
               });
 
               self.route = '#user/settings/' + subView;
+              self.title = title || '';
               publishPageChange('settings-page');
               defer.resolve();
             })
@@ -193,12 +196,13 @@ function (
       }
 
       // request for the widget
-      this.set('UserSettings', settingsPreferencesView('UserSettings', undefined));
+      this.set('UserSettings', settingsPreferencesView('UserSettings', undefined, 'User Settings'));
 
       // request for the widget
-      this.set('UserPreferences', settingsPreferencesView('UserPreferences', 'librarylink'));
+      this.set('UserPreferences', settingsPreferencesView('UserPreferences', 'librarylink', 'Library Link Server'));
 
       this.set('AllLibrariesWidget', function (widget, subView) {
+        var self = this;
         var defer = $.Deferred();
 
         if (redirectIfNotSignedIn()) {
@@ -212,6 +216,7 @@ function (
             app.getWidget('AllLibrariesWidget').done(function (widget) {
               widget.setSubView({ view: subView });
               self.route = '#user/libraries/';
+              self.title = 'User Libraries';
               publishPageChange('libraries-page');
             });
             defer.resolve();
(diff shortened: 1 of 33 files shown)


src/js/wraps/discovery_mediator.js
```diff
@@ -55,15 +52,20 @@ function (
...
-
     if (feedback.query) {
       app.getObject('AppStorage').setCurrentQuery(feedback.query);
       app.getObject('AppStorage').setCurrentNumFound(feedback.numFound);
     } else {
       app.getObject('AppStorage').setCurrentQuery(null);
     }
 
-    this.getPubSub().publish(this.getPubSub().NAVIGATE, 'results-page');
+    var stopNavigation = feedback.query
```
romanchyla: maybe this is the place that needs rethinking; the discovery mediator is listening and handling search start cycles (after the first query successfully returned, it calls results page)

I'm guessing that `__stopNavigation` is there to handle situations this handler picks. What situations are those however?

I'll need to think more about it, my feeling is that if we pass signals to ourselves through query parameters, we might be able to accomplish the same with some reordering or with being more specific to what this handler picks up

src/js/apps/discovery/navigator.js
```diff
@@ -569,31 +572,28 @@ function (
...
+          } else {
+            this.route = '#search/' + data.q.url();
+          }
+          this.replace = false;
+          if (data.page) {
+            data.q.set('__stopNavigation', 1);
+            self.get(data.page).execute(data.page).then(function () {
+              defer.resolve();
+            });
+            return ps.publish(ps.START_SEARCH, data.q);
```
romanchyla: @thostetler  you have mentioned have duplicate searches happening; so if search-page gets called from the discovery mediator, then the search gets initiated also here, could that be the source of double entries?

src/js/wraps/discovery_mediator.js
```diff
@@ -420,33 +422,19 @@ function (
...
-              && app.getWidgetRefCount('Results') >= 1
-          ) {
-            // simply navigate to search results page, widgets are already stocked with data
-            if (app.hasService('Navigator')) {
-              app.getService('Navigator').navigate('results-page', { replace: true });
-              stupidGoAhead = false;
-            }
-          }
-
           if (this.getCurrentPage() !== 'SearchPage' && app.getWidgetRefCount('Results') <= 0) {
```
romanchyla: line 425 has always been a hack, another candidate for examination...

thostetler:
There are a few issues that discuss the history weirdness:
https://github.com/adsabs/bumblebee/issues/1634
https://github.com/adsabs/bumblebee/issues/1527
https://github.com/adsabs/bumblebee/issues/1733

Here is a gif that sort of shows the problem:
history_issue_bbb

Probably part of the problem with this ticket is that I'm trying to fix a few different things at once, but I'll try to explain a little better in the comments

src/js/components/navigator.js
```diff
@@ -74,82 +75,38 @@ function (
...
+          this.router.navigate(route, { trigger: false, replace: replace });
         }
 
         // clear any metadata added to head on the previous page
         $('head').find('meta[data-highwire]').remove();
-        // XXX:rca - this can probably go....anyways, shouldn't be here, is not generic
-        // and set the default title
-        document.title = (transition.title ? transition.title + ' - ' : '') + 'NASA/ADS Search';
-
-        if (!this.globalLinksHandled && Backbone.history.options.pushState) {
```
romanchyla: oh, i'm happy this one is going :+1:

src/js/components/query_mediator.js
```diff
@@ -223,6 +223,15 @@ function (
        * Happens at the beginning of the new search cycle. This is the 'race started' signal
        */
     startSearchCycle: function (apiQuery, senderKey) {
+
+      // If we are passed a blank query, it will fail below this
```
romanchyla: perhaps a matter of taste, i'd prefer to generate an exception if we were given an empty query

src/js/widgets/results/widget.js
```diff
@@ -228,23 +236,6 @@ function (
       }
     },
 
-    updateSidebarsFromUserData: _.debounce(function () {
```
romanchyla: just an observation; if user preference is stored and a page reloaded, it will be hard to set the desired state if there is only toggle

PR #1747, opened 2019-03-18
History issues

romanchyla:
hi @thostetler this is work in progress, please see comments as they appear

solved: https://github.com/adsabs/bumblebee/issues/1733
solved: logo click reloading the application

The history entries are still wonky - I believe it is the interaction between the link handler and the navigator; ideally we would be calling navigator but instead we are calling router -- well, it is hard to not call router, cause the only thing we have are the links; i'm investigating the relationship...

One more thing is in the picture: with pushstate false, who was catching the user acitons? (clicks) -- something must have been doing it before, now we don't give that 'something' the same chance, becuase we handle it and we stop the propagation...I'm going down that route now...

### DIFF
--- a/src/js/apps/discovery/navigator.js
+++ b/src/js/apps/discovery/navigator.js
@@ -589,9 +589,10 @@ function (
           pages = searchPageAlwaysVisible;
         }
 
-
+        var that = this;
         showResultsPage(pages).then(function() {
           self.getPubSub().publish(self.getPubSub().START_SEARCH, data.q);
+          that.route = '#search/' + queryUpdater.clean(data.q).url();
           defer.resolve();
         })
         return defer.promise();
--- a/src/js/components/navigator.js
+++ b/src/js/components/navigator.js
@@ -80,6 +80,8 @@ function (
 
       var self = this;
       var afterNavigation = function() {
+        console.log('afterNavigation', transition, arg1);
+
         // router can communicate directly with navigator to replace url
         var replace = !!((transition.replace || arg1 && arg1.replace));
 
--- a/src/js/components/navigator.js
+++ b/src/js/components/navigator.js
@@ -80,7 +80,6 @@ function (
 
       var self = this;
       var afterNavigation = function() {
-        console.log('afterNavigation', transition, arg1);
 
         // router can communicate directly with navigator to replace url
         var replace = !!((transition.replace || arg1 && arg1.replace));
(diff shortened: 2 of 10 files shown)


romanchyla:
haha, so there were two secret ingredients

1. search-page and results-page were being called at the same time (they both did the same)
2. transition.route needs to be set if a history entry should be created

true, it was arduous to find out, but it makes sense - all the navigation can be controlled from inside navigator

when `that.route` is not set, then no new entry is created (no matter if replace is set or not)

also few minor details -- templates should always use hash tags; we should activate click handling only if pushstate is activate -- I have moved the code

i have tested all history transitions in hash and pushstate modes (using chrome only) by

1. loading page (checking number of entries in the history)
2. doing search (checking ....)
3. doing search (checking....)
4. going to abstract (checking...)

and then navigating back and forth

this issue is fixed too, however the libraries don't expose library id - not sure if it's there: https://github.com/adsabs/bumblebee/issues/1634

also, classic/modern tabs don't work (i.e. I can't start from the classic form), so I didn't try https://github.com/adsabs/bumblebee/issues/1527

however, i suspect it is again something minor :)

src/js/apps/discovery/navigator.js
```diff
@@ -589,9 +589,10 @@ function (
           pages = searchPageAlwaysVisible;
         }
 
-
+        var that = this;
         showResultsPage(pages).then(function() {
           self.getPubSub().publish(self.getPubSub().START_SEARCH, data.q);
+          that.route = '#search/' + queryUpdater.clean(data.q).url();
```
romanchyla: bingo!

src/js/components/navigator.js
```diff
@@ -92,14 +94,44 @@ function (
           // update the History object
           self.router.navigate(
             transition.route,
-            { trigger: true, replace: replace }
+            { trigger: false, replace: replace }
```
romanchyla: bingo!! (to not have duplicate entries)

src/js/wraps/discovery_mediator.js
```diff
@@ -63,7 +63,7 @@ function (
       app.getObject('AppStorage').setCurrentQuery(null);
     }
 
-    this.getPubSub().publish(this.getPubSub().NAVIGATE, 'results-page');
+    //this.getPubSub().publish(this.getPubSub().NAVIGATE, 'results-page');
```
romanchyla: bingo!!! (to not have duplicate search events)

likely this place needs to be examined more, the code few lines below was there for some transition -- this is from before all was tidily stuffed into navigator

thostetler:
@romanchyla, this is looking pretty good, I still see some of the other issues, I'll remove the debug stuff and merge, and try to see what from my other PR is still necessary.

romanchyla:
Cool, thanks!

PR #1755, opened 2019-03-20
Update Global Link Handler and Remove Extras

thostetler:
As per Roman's suggestion, moving the global handler to run _after_ `history.start()`

### DIFF
--- a/src/js/mixins/discovery_bootstrap.js
+++ b/src/js/mixins/discovery_bootstrap.js
@@ -20,6 +20,68 @@ function (
   HandleBars,
   ApiTargets
 ) {
+
+  var startGlobalHandler = function () {
+    var routes = [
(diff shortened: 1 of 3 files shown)


src/js/mixins/discovery_bootstrap.js
```diff
@@ -219,6 +281,9 @@ function (
         }, conf && conf.routerConf);
         Backbone.history.start(newConf);
 
+        // apply a global link handler for push state (after router is started)
```
romanchyla: just sayin' - we could have used simpler case of detecting pushstate before callng `startGlobalHandler()`  -- as it is now, it allows to enable/disable pushstate in realtime; i guess might be useful (?) so i'm merging as it is

PR #1760, opened 2019-03-20
Fix 36

romanchyla:
closes item 36:

Priority] (AA) Starting from the classic style search, enter an author (e.g. “accomazzi”) and select refereed returns just refereed papers (correctly), but removing the property filter from the results list causes the results to still show just refereed papers (although the breadcrumb for property is gone).  It appears that the refereed filter is still encoded in the URL, but not reflected in the UI.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -45,7 +45,6 @@ http{
 
               expires 6M;
 
-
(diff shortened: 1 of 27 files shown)


thostetler:
LGTM

PR #1761, opened 2019-03-21
Fix A few Routes

thostetler:
This fixes the issue with the library not showing correct route
Also the issue with abstract page export not loading properly
Update a few tests to get things working

### DIFF
--- a/src/js/apps/discovery/navigator.js
+++ b/src/js/apps/discovery/navigator.js
@@ -234,14 +234,15 @@ function (
         var that = this;
         // this is NOT navigable from outside, so library already has data
         // only setting a nav event to hide previous widgets
-        app.getWidget('IndividualLibraryWidget').done(function (widget) {
-          widget.setSubView({ subView: 'admin' });
-          app.getObject('MasterPageManager').show('LibrariesPage',
+        app.getObject('MasterPageManager').show('LibrariesPage',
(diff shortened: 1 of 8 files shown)


src/js/apps/discovery/navigator.js
```diff
@@ -208,77 +213,76 @@ function (
...
+              // show the pages
+              app.getObject('MasterPageManager').show(
+                pub ? 'PublicLibrariesPage' : 'LibrariesPage',
+                pub
+                ? ['IndividualLibraryWidget', 'LibraryListWidget']
+                : ['IndividualLibraryWidget', 'LibraryListWidget', 'UserNavbarWidget']
+              ).then(function () {
+                if (pub) publishPageChange('libraries-page');
+
+                // TODO: unable to find the solution, these widgets are getting cleaned up before being shown
```
romanchyla: hmm, i'll have to look at it carefully...

romanchyla:
@thostetler please have a look at my latest changes; not sure what the bug was - i've reworked the order of events and i can load/reload bookmarked library with ID now

src/js/widgets/library_list/widget.js
```diff
@@ -175,6 +175,7 @@ define([
         'libraryID': data.id,
         'editRecords': data.editRecords
       });
+      this.dispatchRequest();
```
thostetler: gahh! yeah this makes sense, I didn't think about moving when the request is made.

PR #1778, opened 2019-04-02
Re-added 'remove' operation to query updater

romanchyla:


romanchyla:
@thostetler i should have added tests, really, but in the interest of time, not to block any release if you wanted to use this feature, i'm sending it now

thostetler:
This looks good, I added a PR to make the graph facet work with these changes.  I'll merge after Travis does its thing

PR #1786, opened 2019-04-03
Fix search url not updating issue

thostetler:
Since we removed the navigation event from START_SEARCH, the calls to it in the results page widgets no longer update the url.  Swapping it out for a navigation event (which also triggers search) works.

Added a little more robust solution for handling the replacement of query params updated via the graph facets.  I think this makes these widgets much more useful.  One issue is that they don't really like it if the the query changes (via manually updating in search bar).


thostetler:
Thanks @romanchyla, I've updated the PR.  I can confirm, afaict we only ever add a single history entry for subsequent (identical) searches.  Running that navigation event 4-5 times in a row only runs the search once.

PR #1792, opened 2019-04-04
make alt ids work again

thostetler:
Swap out `bibcode:` style queries for using `identifier:` for all abstract pages, replaces alternate ids in the url with bibcode.  Not positive if this is a definitive fix, are there cases that this will point the wrong document? we will need to test this out more thoroughly.

for example:
`/abs/doi%3A10.1117%2F12.2314185/abstract` -> `/abs/2018SPIE10699E..02H/abstract`

Also abstract page should show a proper error if the id is missing now.
Also updated the resources widget to match the way ids are handled in the other widgets.


thostetler:
@romanchyla merging for now, but please review when you get a chance.

romanchyla:
Looks good Tim

PR #1833, opened 2019-05-12
Add second order operator support

thostetler:
Adds support for second order operations.

second_order_ops

### DIFF
--- a/src/js/components/second_order_controller.js
+++ b/src/js/components/second_order_controller.js
@@ -245,7 +245,10 @@ function (
         }
 
         // replace the current query with our operator
-        const newQuery = new ApiQuery({ q: `${ field }(docs(${ qid }))` });
+        const newQuery = new ApiQuery({ 
+          q: `${ field }(docs(${ qid }))`,
+          sort: 'score desc'
+        });
         ps.publish(ps.NAVIGATE, 'search-page', { q: newQuery });
 
         this.submitAnalyticsEvent(field);
--- a/src/js/components/second_order_controller.js
+++ b/src/js/components/second_order_controller.js
@@ -1,0 +1,146 @@
+define([
+  'js/components/generic_module',
+  'js/mixins/dependon',
+  'js/components/api_query',
+  'js/components/api_request',
+  'js/components/api_targets',
+  'analytics'
+],
+function (
+  GenericModule,
+  Dependon,
+  ApiQuery,
+  ApiRequest,
+  ApiTargets,
+  analytics
+) {
+
+  /**
+   * Triggered via pubsub event, this will take a set of identifiers
+   * and generate a bigquery id, then replace the current query
+   * with a specified field
+   *
+   * example query: `similar(docs(99999))`
+   */
+  const SecondOrderController = GenericModule.extend({
+    activate: function (beehive) {
+      this.setBeeHive(beehive.getHardenedInstance());
+      const ps = this.getBeeHive().getService('PubSub');
+      ps.subscribe(ps.CUSTOM_EVENT, _.bind(this.onCustomEvent, this));
+    },
+
+    /**
+     * Expecting an event in the following format
+     * `second-order-search/{ field }`
+     * then call the transform method
+     */
+    onCustomEvent: function (event) {
+      if (event.startsWith('second-order-search/')) {
+        this.transform.apply(this, [event.split('/')[1]].concat(_.rest(arguments)));
+      }
+    },
+
+    /**
+     * Grab the list of currently selected papers from app storage
+     */
+    getSelectedIds: function () {
+      const storage = this.getBeeHive().getObject('AppStorage');
+      if (storage && storage.getSelectedPapers) {
+        return storage.getSelectedPapers() || [];
+      }
+    },
+
+    /**
+     * Grab the qid from vault by sending our list of bibcodes
+     * returning a promise
+     */
+    getBigQueryResponse: function (ids) {
+      const ps = this.getPubSub();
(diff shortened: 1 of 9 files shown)


src/js/components/second_order_controller.js
```diff
@@ -0,0 +1,269 @@
...
+
+      const options = _.defaults({}, opts, {
+        onlySelected: false
+      });
+
+      // get the selected records from appStorage
+      const selectedIds = this.getSelectedIds();
+
+      if (selectedIds.length === 0 || !options.onlySelected) {
+        this.getIdsFromCurrentQuery()
```
romanchyla: so this could spell problems; it downloads big number of ids, then creates an entry in the database, then behind the curtain this will be loaded into another query, sent to solr.... (bigquery stuff is really inefficient)

instead, the preferred method imho, is to wrap the query that produces those ids

```qString = this.serializeQuery(this.getCurrentQuery().clone())```

which would produce something like:

`trending(star AND author_facet:"0/Accomazzi")`

note, that the docs operator is not necessary here; serialization of the query will not be exactly straightforward because filters are scattered around in different fields, but it is not too hard either - everything can be AND'ed)

only if a bigquery gets added into the mix, do we need to use `docs(....)`

e.g. somehting like 

`trending(star AND author_facet:"0/Accomazzi" AND docs(qid))`

but i think we don not have such a usecase yet

Comment (romanchyla) [reply src/js/components/second_order_controller.js]: btw: one other advantage to this method is that the user can (and will be able to modify) operators applied on top of another operator

for example if they did `co-reads` and then citations

`citations(trending(.....))`

with a query like these, it is possible to parse it, go back to previous step

Comment (thostetler) [reply src/js/components/second_order_controller.js]: I think I understand.  Do you think that we should remove the ability to do perform this on a full result for now? and only allow it if there have been some records selected?

I could then work on serializing the query like you mentioned as part of an update to this?

Comment (marblestation) [reply src/js/components/second_order_controller.js]: @romanchyla we discussed in person and in slack this proposal already, I see the benefits but @aaccomazzi does prefer to fetch the first N results and store them in a database so that it works similarly to the visualization tools in the explore menu. We need @aaccomazzi to be on board to change direction.

Comment (aaccomazzi) [reply src/js/components/second_order_controller.js]: We had this discussion and we made the decision to go with docs() because it works across all use cases up to N records, is consistent with the other filtering done in visualizations, and can be iterated over.   For the power users who really to do this at scale will need to learn how to formulate a query and wrap it with the proper second level operator.  Thanks @romanchyla for the additional option but I think we should stick with the original design until we find use cases that make these changes necessary.

**Comment by romanchyla** [comment]
Ok, I would like to call all parties to show their votes and why. This
discussion was had but as far as I registered, the decision was yours
@aaccomazzi. I have presented arguments why it is a bad idea (it's actually
very bad idea in my opinion) so now I need to make sure that people are not
falling in line just because decision was made from the top. We are getting
more and more into the pattern where decisions are made this way and I have
seen it before in one particular project, and the product they deliver is
not superior to ADS (who knows not which project I have in mind, I can
explain in person)

**Comment by aaccomazzi** [comment]
Sorry you feel excluded, but it wasn't just my diktat.  One thing to keep in mind is that the two proposed solutions will yield different results when the result set is above N records (I think we are setting N to 6,000 in this case).  One reason why I prefer grabbing the top N results and using docs() on them is that it allows the user to control the list by querying, filtering, ranking, before applying second-order operations on the most relevant ones.  Taking the same second-order operator on the entire data set will provide a different result.  I think it's good to have the user in control of this choice.

There's also the issue of what to do when the result set do not easily fit in `q` (e.g. data coming from a previous bigquery from a library, or visualization filter).  Is there really a solution that works across the board in a robust way?  

Finally, putting facets in the query will likely be confusing to the users -- we never document them as searchable fields and they have an awkward syntax, are we sure we want to expose them this way?
```
((author_facet_hier:"0/Accomazzi, A") OR author_facet_hier:"1/Kurtz, M/Kurtz, Michael J" OR author_facet_hier:"1/Kurtz, M/Kurtz, M  J")
```
Trying to translate them to the regular search field seems overkill to me...

romanchyla:
@aaccomazzi i do not need to feel like i was excluded, i simply **was** excluded. The discussion didn't conclude, decision was made - given the comments above from @marblestation I feel compelled to conclude that there was one agent who made the decision. So that's one problem (the important one).

But easily resolved, other people who have input should voice their approval/disapproval here, and I'll see that I'm alone in voicing the objections and it is wiser to give up. But they should well be aware, that this search operation, as implemented is 10-100x slower, makes the UI slow, the search slow, and from the technical perspective is just simply is terribly inferior - there is no other way to put it.

You say you want users to maintain control, but we have the query language and operators to do that - `trending(topn(<number>, <query>))` should then be your first choice then (but nowhere do I see an option to let users control how many results are included, in the present implementation; it seems to be an afterthought)

The argument about it being confusing to the users goes both ways: it can be confusing to some, to some others it will be confusing that you are making *arbitrary* choice of including only first X results where the interface suggests that the operation is applied on the result of the previous search (all of it).

And this is my pet peeve: this is a **search/explore** operation; so it should be treated like a search. I have heard argument that we want to make it consistent with the visualizations, but that is simply wrong. Visualizations are end-states, where one tries to compress and present results of the search operation and relations in one glance. The search is exploration, but the implementation of picking N results (and wrapping them into docs()) makes any exploration impossible; it is a final set, frozen in time, decided by some arbitrary criteria because it seemed right.

Given our propensity not to take back features once implemented, it is therefore important to get this right the first time.

PR #1870, opened 2019-06-21
Remove sw workbox and append release version to js requests

marblestation:
Regarding the service worker (sw):
- it may allow us to reload BBB on the background when there are new files, this would be positive but
- it messes with the browser cache and the server side cache headers
- some browsers do not behave as expected when the user clears the cache and hard reloads
- it makes a fresh load of BBB 2 times slower
- it does not make BBB load faster when things were in cache

The cache issues we experienced before implementing the service worker may be due to an old nginx configuration and the sw does not seem to help. This PR removes the sw and recovers the addition of versions as parameters after every requested `.js`, not because it is needed for avoid caching problems but because it can be useful for logging purposes.


src/index.html
```diff
@@ -94,11 +94,16 @@
...
+
         // define waitSeconds above require script tag
         // to override the default, until main.js loads
         //see https://github.com/yeoman/yeoman/issues/1051
         window.require = {
-            waitSeconds: 30
+            waitSeconds: 30,
+            // this will append a query string to requests for resources, uses the current	
+            // version for production and the current time for development	
+            urlArgs: `v=${ APP_VERSION.length > 0 ? APP_VERSION : (new Date()).getTime() }`,
```
thostetler: Should change this to use normal string concatenation since these template strings will break on older browsers.

Any of the other files this would be fine, since we transpile, but with index.html it will run as is.

thostetler: `'v=' + APP_VERSION.length > 0 ? APP_VERSION : (new Date()).getTime()`

PR #1931, opened 2019-09-25
Provide data link targets when provider is not in built-in list

aaccomazzi:
This provides a sensible defaults when we run into a data link which is not already configured in the LINK_TYPES data structure.


thostetler:
@aaccomazzi you have a small fix in the tests.  

I updated some formatting issues, would you like me to add a commit to fix the test or are you working on it? 

Thanks!

aaccomazzi:
Please go ahead with the commit, Thanks!

PR #1932, opened 2019-10-07
update loading messages

thostetler:
loadingscreen-1
js-required

Update the messages on loading to have links to ADS Core
Update the styling a bit to make it behave a little better on small devices.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -50,7 +50,7 @@
           Sorry, but ADS requires JavaScript to be enabled.
         </p>
         <p class="pull-right">
-          <a href="//ui.adsabs.harvard.edu" class="btn btn-info">Reload</a>
+          <a href="/core/always" class="btn btn-info">Try ADS Core</a>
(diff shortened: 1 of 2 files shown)


marblestation:
The original requirement was to add something like:

```
Loading ADS | <a href="https://ui.adsabs.harvard.edu/core">Load basic HTML</a> (for slow connections/low resources)
```

Without saying `ADS Core` for the reasons we mentioned in slack and in meetings. To build on top of this proposal, I would:

- Remove "Loading is taking a while..."
- Convert the current button to the text I added above
- Make sure the message always appears like Gmail does

For the javascript message, I would change it to:

```
Sorry, but the fully featured ADS requires JavaScript to be enabled. Alternatively, <a href="https://ui.adsabs.harvard.edu/core">load basic HTML</a> (for slow connections/low resources).
```

And I would remove the button which mentions ADS Core.

PR #1970, opened 2019-11-22
Changes to the default sort

romanchyla:
This change is not controversy free; but it would make search use scoring unless some fielded search is present (besides fielded searches that should always be evaluated based on the score; such as fulltext or aff search)

a candidate for A/B testing

### DIFF
--- a/src/js/widgets/search_bar/search_bar_widget.js
+++ b/src/js/widgets/search_bar/search_bar_widget.js
@@ -830,31 +830,58 @@ function (
 
       // make sure not to override an explicit sort if there is one
       if (!query.has('sort')) {
-        var queryVal = query.get('q')[0];
+        var queryVal = query.get('q')[0].trim();
 
-        // citations operator should be sorted by pubdate, so it isn't added here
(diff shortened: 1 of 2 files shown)


thostetler:
@aaccomazzi Do we want to have a discussion on these changes? or is this something we'd like to merge first and check out for a few days to decide?

aaccomazzi:
Not ready for prime time, we should revert, but some thoughts about the PR:

1. It would be a good time to review the default scoring for each operators (e.g. should references() still use 'first_author_norm asc'?
2. don't quite understand the 'score asc' case, should be 'desc' for sure

romanchyla:
oh, i think both the dev and the reviewer were asleep at the wheel :-)

romanchyla:
my plan is to decide on this feature using A/B testing; measuring how often people switch the default sort mode; we don't seem to be able to make a sensible decision without gathering data - i'm ready to claim that if we find no statistically significant change if we serve both versions; then it makes no difference to users and therefore the score is a better choice; if we find statistically significant difference, then obviously the one that people use more should become the default

the missing part right now really is the tooling for A/B testing....

PR #2024, opened 2020-04-10
AB Testing

romanchyla:
As can be seen here, the changes are kinda minimal - I'm however expecting experiments.js to evolve quite a bit. 
As for the location of the analytics.js -- when retrieved from /analytics/analytics.js then it is going to use our proxy; which will get deployed soon (it already is on dev) -- but multiple targets suggest that require will proceed onto loading the second target if the first is not avaialbe (i haven't tested it though)


src/js/components/experiments.js
```diff
@@ -1,52 +1,47 @@
...
+      this.toggleOptimize();
+      // var self = this;
+      // setTimeout(function() {self.toggleOptimize(), 1000}); // won't be necessary if I knew what event to listen to...
+    },
 
-    return Experiments;
-});
\ No newline at end of file
+    toggleOptimize: function() {
+      const dataLayer = window.dataLayer;
```
thostetler, line 30: Assuming `dataLayer` is a global property.  If not we can amend this then.

PR #2033, opened 2020-04-30
Ab2

romanchyla:


src/config/common.config.js
```diff
@@ -42,19 +42,35 @@ define([], function() {
     },
   });
 
-  window.GoogleAnalyticsObject = '__ga__';
-
   require(['config/discovery.vars'], function(config) {
+    
+    // rca: not sure why the ganalytics is loaded here instead of inside analytics.js
```
romanchyla, line 47: @thostetler just out of curiosity; this block could live in the analytics

src/js/apps/discovery/main.js
```diff
@@ -63,6 +63,12 @@ define(['config/discovery.config', 'module'], function(config, module) {
         updateProgress(100);
 
         app.onBootstrap(data);
+
+        var dynConf = app.getObject('DynamicConfig');
```
romanchyla, line 67: i have moved this higher up so that experimentns can subscribe; i haven't noticed any bad side effects

src/js/components/analytics.js
```diff
@@ -70,33 +70,34 @@ define(['underscore', 'jquery'], function(_, $) {
...
-        },
-      ]);
-      setTimeout(() => $dd.resolve(), 1000);
-      return $dd.promise();
-    } catch (e) {
-      console.info('google analytics event not tracked');
+  var buffer = [];
+  var gaName = window.GoogleAnalyticsObject || 'ga';
+
+  var cleanBuffer = function() {
```
romanchyla, line 76: @thostetler i did rework analytics little bit; the main changes are:

1. no wrapper around the call if analytics is present
2. have a simple buffer for situations when g.js wasn't initialized yet

src/js/react/BumblebeeWidget.js
```diff
@@ -53,6 +67,16 @@ define([
...
       });
+
+      if (this.view._store) {
+        this._store = this.view._store;
+      }
+    },
+    dispatch({ type, ...args }) {
+      this._store.dispatch({ type, ...args });
+    },
+    getState() {
```
romanchyla, line 78: i'm not familiar with React/redux - getState() i got that working, but i was missing setState()

by setting variables directly into the state, i did change it, but that didn't result in changes in behaviour of the widget

src/js/wraps/landing_page_manager/landing_page_manager.js
```diff
@@ -63,6 +65,17 @@ define([
       }
     },
 
+    onWidgetSelected(child, ev, attrs) {
+      if (attrs.idAttribute === 'SearchWidget') {
+        const Recommender = new RecommenderWidget();
```
romanchyla, line 70: @thostetler this place will need some more thinking; i have temporarily solved it this way so that on the page the experiment can do the following

```
var exp = bbb.getController('Experiments');
if (exp) {
  exp.subscribe("NAVIGATE", 
     function(id, data) {
       if (id === 'SearchWidget') {
         bbb.getWidget('SearchWidget').then(function(w) {
           window.r = w.recommender;
           var state = w.recommender.getState();
           // activate recommendations
           if (!state.doJustOnce) {
              var actions = require('js/react/Recommender/actions');
              w.recommender.dispatch(actions.setTab(1));
              state.doJustOnce = true;
              console.log("Congratulations, you have been selected to be a guinea pig!");
           }
         })
       }
     }
  )
}```

what gave me most resistance was the getTab() action; in the end i ended up importing it - but not sure if that is the recommended way

```require('js/react/Recommender/actions');```

i was secretly hoping this would work

`state['state'] = 1`

but while it did set the state, it also didn't trigger the action (and I couldn't call the onSelectedWidget(1) because that would trigger analytics call -- we want that only when users click)

File: src/js/react/Recommender/actions.js [hunk 6]
```diff
@@ -0,0 +1,69 @@
...
+      scope: actions.GET_RECOMMENDATIONS,
+      options: {
+        type: 'POST',
+        data: {
+          function: func,
+          sort,
+          num_docs: numDocs,
+          cutoff_days: cutOffDays,
+          top_n_reads: topNReads,
+          reader: '3138c73ffff34a0f'
```
romanchyla, line 26: @thostetler the reader value is hardcoded here -- but that's only for testing; in the real prod we'd not need it -- however, for testing it would be extremely helpful to be able to modify parameters that will be used for search

could you think of a way to do this elegantly? (i.e. we should not try to design the widget around AB experiments, but it would be nice to be able to modify those values  -- maybe get them from the state?)

the same comment could apply to the target url; i know there is an endpoint for setting up the target url, though that too seems bit arbitrary (as if it was built for AB); thoughts?

romanchyla: btw: i did hardcoded that value in there, in case that was not obvious...

src/js/react/BumblebeeWidget.js
thostetler: Hey Roman, yeah you can't set state directly in redux you have to do it by dispatches.  
So for example, if you wanted to update the tab from 1 to 2 you would do something like:

`Recommender.dispatch({ type: 'SET_TAB', payload: 2 });`

I would suggest grabbing the `actionCreator` from https://github.com/romanchyla/bumblebee/blob/ab2/src/js/react/Recommender/actions.js#L58
then you could do:

`Recommender.dispatch(setTab(2));`

then the corresponding trigger will run in the reducer:
https://github.com/romanchyla/bumblebee/blob/ab2/src/js/react/Recommender/reducer.js#L55

Its a little verbose, but this allows you to guarantee the store is never updated in a weird way, it is always up to date.

src/js/react/Recommender/actions.js
thostetler: Yep, I agree I think we should move these up to the redux store then we can just set them up when you initialize the component.  I could also an initial store to be set in the constructor, but you then would probably have to know the structure of the store ahead of time.
i.e:
```js
new Recommender({ 
  store: { 
    tab: 1, 
    queryParams: { 
      reader: '3138c73ffff34a0f'
    }
  }
});
```
or 
```js
const r = Recommender();
r.dispatch(setTab(1));
r.dispatch(setQueryParams({ reader: '3138c73ffff34a0f' });
```

src/js/wraps/landing_page_manager/landing_page_manager.js
thostetler: yeah I agree, there should be a better way of loading things in.  Yeah how you're doing dispatching is correct, at least with redux that is the way it has to be done.  

Yeah redux only starts its cycle when an action is dispatched, so modifying the state directly doesn't do anything.  In the future we could potentially switch to something like mobx or some other library which uses observers on the objects, but in the end all of them strive for immutability one way or another.

src/js/react/BumblebeeWidget.js
thostetler: Yeah with Redux you have to dispatch actions, it doesn't have a way of setting the state directly...

src/config/common.config.js
thostetler: Yep I agree, as far as I know the GA init stuff has lived in the requirejs config callback.  I think, like you mention, that it is because it will be available as soon as possible.  
I can make a separate PR then to move it into analytics to clean this file up a bit.

src/js/react/Recommender/components/RecommendedList.jsx.js
```diff
@@ -0,0 +1,165 @@
...
+        dispatch(getRecommendations());
+      }
+    }, [docs]);
+
+    const onPaperSelect = ({ bibcode }, index) => {
+      dispatch( // toto: how to get state?
+        emitAnalytics([
+          'send',
+          'event',
+          'interaction.recommendation', // need to add here '.' + state.queryParams['function']
```
romanchyla, line 99: hey @thostetler what would be the nicest way to get access to the state here? I see the execution of middleware() is needed, i contemplated senting the data from actions too, but it was heavy handed. The idea here is to send `interaction.recommendation.similar` event to be able to distinguish different algs/results

thostetler: Right above there I'm using the `useSelector` hook to get access to some of the state.  We could update the selector just above that to include `queryParams`: 
```
const selector = (state) => {
  return {
    getRecommendationsRequest: state.requests.GET_RECOMMENDATIONS,
    getDocsRequest: state.requests.GET_DOCS,
    docs: state.docs,
    queryParams: state.queryParams
  };
};
```

and then update the variables returned by the hook:
```
const { getRecommendationsRequest, getDocsRequest, docs, queryParams } = useSelector(
  selector
);
```
and then finally you can do something like:
```
const onPaperSelect = ({ bibcode }, index) => {
  const { function: func } = queryParams;
  dispatch(
    emitAnalytics([
      'send',
      'event',
      `interaction.recommendation.${ func }`,     
      index,
      { bibcode },
    ]);
  );
};
```

src/js/react/Recommender/components/RecommendedList.jsx.js
```diff
@@ -0,0 +1,200 @@
...
+      getRecommendationsRequest,
+      getDocsRequest,
+      docs,
+      queryParams,
+      userName
+    } = useSelector(selector);
+
+    React.useEffect(() => {
+      if ((executed + 12*60*60*1000) < Date.now()) {
+        // the hook gets called too many times even with [docs] in the args to useEffect
```
thostetler, line 105: When the `useEffect` is written like this without the second argument stating the dependencies it will run every render.  Not sure looking at this what would need to be true/not be true for the `getRecommendations()` dispatch to run

romanchyla:
Yes, I know, but I have found the same to be true when used with docs (or
the opposite, I don't recall, when docreq.status) was used. Even with docs,
it was being evoked twice

PR #2065, opened 2020-08-20
classic-form: add support for query params

thostetler:
Adds support for the following query params:

`collection` and `property`

can be used like this:

`/classic-form?collection=general,physics,astronomy&property=refereed,article`

can also use shorthand:

`/classic-form?c=g,p,a&p=r,a`

or combination:

`/classic-form?collection=g,a&property=a`

### DIFF
--- a/src/js/apps/discovery/navigator.js
+++ b/src/js/apps/discovery/navigator.js
@@ -175,15 +175,16 @@ define([
         return defer.promise();
       });
 
-      this.set('ClassicSearchForm', function() {
+      this.set('ClassicSearchForm', function(widgetName, { query }) {
         var defer = $.Deferred();
         var that = this;
         app
           .getObject('MasterPageManager')
-          .show('LandingPage', ['ClassicSearchForm'])
+          .show('LandingPage', [widgetName])
           .then(function() {
             app.getWidget('LandingPage').done(function(widget) {
-              widget.setActive('ClassicSearchForm');
+              widget.setActive(widgetName);
+              widget.widgets[widgetName].applyQueryParams(query);
             });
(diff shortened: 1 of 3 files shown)


golnazads:
> Adds support for the following query params:
> 
> `collection` and `property`
> 
> can be used like this:
> 
> `/classic-form?collection=general,physics,astronomy&property=refereed,article`
> 
> can also use shorthand:
> 
> `/classic-form?c=g,p,a&p=r,a`
> 
> or combination:
> 
> `/classic-form?collection=g,a&property=a`

@golnazads

golnazads:
@thostetler 
all other parameters coming from tugboat have colon in between the field and value, you have equal sign here. Should tugboat continue with colon? thank you.

thostetler:
@golnazads should always be `=` for URL params.  Can you paste me an example of one with colons?

golnazads:
for example year:2001 that is how year is translated and sent to BBB.
Actually here are unittest and this one is for pubdate
https://github.com/adsabs/tugboat/blob/master/tugboat/tests/tests_unit/test_search_redirect.py#L200

thostetler:
@golnazads 
`...tEqual('q=' + urllib.quote('pubdate:[1990-01 TO 1991-12]') + '&sort=' + urllib.quote('dat...` the part with the colon is inside the `q=` param, I could change away from commas to delimit, but we'd have to keep the `=` for it to get parsed properly

golnazads:
After our discussion in standup it seems that it should not be part of the
tugboat. But let me hear back from Alberto for the final decision.

But just for my knowledge, these two parameters do not go inside the query,
right?

thank you so much.

PR #2112, opened 2021-01-29
Improve highlights ux, add star to query

thostetler:
Adds `*` to `hl.fl` field for highlights
Adds a `no highlights` message when none are returned

image

### DIFF
--- a/src/js/widgets/list_of_things/widget.js
+++ b/src/js/widgets/list_of_things/widget.js
@@ -268,7 +268,7 @@ define([
         // make sure the new docs close highlights if they aren't there
         var newDocs = _.map(docs, function(d) {
           return _.extend(d, {
-            showHighlights: !_.isEmpty(d.highlights),
+            showHighlights: typeof d.highlights !== 'undefined',
             showCheckbox: !!self.model.get('showCheckboxes'),
           });
         });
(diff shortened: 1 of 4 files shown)


src/js/widgets/list_of_things/widget.js
```diff
@@ -606,7 +606,7 @@ define([
           var q = this.model.get('currentQuery').clone();
           q.set({
             hl: 'true',
-            'hl.fl': 'title,abstract,body,ack',
+            'hl.fl': 'title,abstract,body,ack,*',
```
romanchyla, line 609: i think it is actually better to just say: `'hl.fl': '*'`

romanchyla (approved):

src/js/widgets/list_of_things/widget.js
thostetler: I'll check it out, but doing '*' returned an empty param and causes an error in the response parser.  I didn't look into it at all though, just figured they were equivalent

thostetler: doing `hl.fl: '*'` 
image

gets an empty response: 
image

That is causing an error.  Anything I can do about that? is that correct behavior

thostetler:
I'm going to merge this, we can circle back around and make the change to `*` later.  Assuming the changes are equivalent.

PR #2127, opened 2021-03-12
Fix submenu overlap

shinyichen:
Adjusted navigation submenu so that they don't overlap of top-level menu in mobile screen.


thostetler:
Nice! easy enough to fix LGTM

PR #2128, opened 2021-03-12
Send analytics event

shinyichen:
Send analytics event for full article links

### DIFF
--- a/src/js/widgets/abstract/widget.js
+++ b/src/js/widgets/abstract/widget.js
@@ -17,6 +17,7 @@ define([
   'mathjax',
   'bootstrap',
   'utils',
+  'analytics',
 ], function(
   Marionette,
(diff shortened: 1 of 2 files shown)


src/js/widgets/abstract/widget.js
```diff
@@ -58,6 +60,7 @@ define([
     },
 
     parse: function(doc, maxAuthors = MAX_AUTHORS) {
+      console.log(doc);
```
thostetler, line 63: we should remove this line

thostetler (changes requested):

src/js/widgets/abstract/widget.js
shinyichen: Removed

thostetler:
Looks good, thanks!

PR #2129, opened 2021-03-12
Force EST timezone for tests

thostetler:
Use puppeteer's built-in timezone setting to force EST during testing


shinyichen (approved):
Yes that resolved it!

PR #2134, opened 2021-03-22
Hide mouse over dropdown menu if mobile or tablet devices

shinyichen:


src/js/widgets/list_of_things/model.js
```diff
@@ -139,6 +140,19 @@ define([
...
+      const devices = [
+        /Android/i,
+        /webOS/i,
+        /iPhone/i,
+        /iPad/i,
+        /iPod/i,
+        /BlackBerry/i,
+        /Windows Phone/i,
+      ];
+      var isMobileOrTablet = devices.some((d) =>
```
thostetler, line 152: This certainly works.  However we already have a library `bowser` that could parse the user agent and provide more coverage.
There is a utility: https://github.com/adsabs/bumblebee/blob/a1723ff2c0c17940cac7b4da17b50a1ee1d838a9/src/js/utils.js#L56

or docs here: https://lancedikson.github.io/bowser/docs/index.html

thostetler (approved):

PR #2137, opened 2021-03-25
Fixed library issues

shinyichen:
This PR addresses the following three issues of library view:
* When user viewing a public library who is not a collaborator, the edit buttons says “You have none privileges", but allows user to click on “View editing options” and add/remove papers. 
  * In this case, when user has no permission, this edit button is removed.
* When selecting records, the “delete” option appears but the user cannot actually delete any record.
  * Delete button is not shown if user has no edit permission
* "View library in results page" button is not shown unless user has edit permission
  * Show "View library in results page" button regardless of permission


thostetler (approved):
Looks good, there is a minor conflict -- can merge any time then

PR #2139, opened 2021-04-07
Minor layout changes

shinyichen:
* Removed side margins when in single column view (search result & abstract)
* Fit metadata into one row in single column view

Screen Shot 2021-04-07 at 11 08 30 AM


thostetler:
LGTM

PR #2148, opened 2021-04-22
"Add to library" widget in single column view

shinyichen:
Make "Add to library" widget accessible in single column view.
* Provide a button to slide the widget out 
* Make it stick to top when scrolled to

Screen Shot 2021-04-22 at 12 10 55 PM

### DIFF
--- a/src/js/apps/discovery/main.js
+++ b/src/js/apps/discovery/main.js
@@ -171,6 +171,17 @@ define(['config/discovery.config', 'module'], function(config, module) {
             }
           });
 
+          $('body').on('click', '#results-actions-toggle', function() {
+            $('#query-info-container').toggleClass('show');
+            if ($('#query-info-container').hasClass('show')) {
+              $('#results-actions-toggle').html(
+                '<i class="fa fa-times" alt="close actions"></i> Actions'
+              );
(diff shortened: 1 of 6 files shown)


src/js/apps/discovery/main.js
```diff
@@ -171,6 +171,19 @@ define(['config/discovery.config', 'module'], function(config, module) {
             }
           });
 
+          $('body').on('click', '#results-actions-toggle', function() {
+            $('#query-info-container').toggleClass('show');
+            if ($('#query-info-container').hasClass('show')) {
+              $('#results-actions-toggle').html(
+                '<i class="fa fa-times" alt="close actions"></i> Actions'
```
thostetler, line 178: For these `<i>` elements, I think I would use `title="..."` instead of `alt`
and add `aria-hidden="true"` so that screen readers will skip it.  I think that that should work for screen readers, since I believe they'll still pull in that `title` information

thostetler:
Small comment to review, feel free to merge when ready

src/js/apps/discovery/main.js
shinyichen: Done

PR #2152, opened 2021-04-24
Split application settings into search and export

shinyichen:
Screen Shot 2021-04-23 at 5 20 43 PM

Screen Shot 2021-04-23 at 5 20 51 PM


thostetler:
nice work splitting these, looks like it was a pain to go through

PR #2167, opened 2021-05-21
Fixed bug: Results page sidebar disappears after logging in

shinyichen:


thostetler (approved):
Definitely a bug I added.  Nice work tracking it down! 👍

PR #2168, opened 2021-05-22
Disable action button when user is not logged in

shinyichen:
When user is not logged in there, no add to library action can be performed. Disable the button

Screen Shot 2021-05-21 at 6 08 40 PM

### DIFF
--- a/src/js/page_managers/three_column_view.js
+++ b/src/js/page_managers/three_column_view.js
@@ -125,12 +125,11 @@ define([
       $leftCol = this.$('#results-left-column');
       $rightCol = this.$('#results-right-column');
       $middleCol = this.$('#results-middle-column');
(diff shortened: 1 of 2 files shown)


src/js/page_managers/three_column_view.js
```diff
@@ -125,12 +125,11 @@ define([
       $leftCol = this.$('#results-left-column');
       $rightCol = this.$('#results-right-column');
       $middleCol = this.$('#results-middle-column');
-      $action = this.$('#results-actions-toggle');
 
       _.each(
         [
           ['left', leftState, $leftCol],
-          ['right', rightState, $rightCol, $action],
```
shinyichen, line 133: The action button visibility is now handled by media query based on screen size, these code are no longer needed

thostetler (approved):

PR #2175, opened 2021-06-04
Download CSV data

shinyichen:
For each metrics and histogram, provided download button to download corresponding CSV data.
Screen Shot 2021-06-21 at 3 30 45 PM
Screen Shot 2021-06-21 at 3 31 29 PM

### DIFF
--- a/src/js/widgets/bubble_chart/templates/container.html
+++ b/src/js/widgets/bubble_chart/templates/container.html
@@ -20,7 +20,7 @@
     viewBox="0 0 1160 550"
     preserveAspectRatio="xMinYMin meet"
   ></svg>
-  <i class="fa fa-download fa-2x download s-download" alt="download data" title="download data"></i>
+  <i class="fa fa-download fa-2x download s-download s-download-lr" alt="download data" title="download data"></i>
   {{/if}}
 </div>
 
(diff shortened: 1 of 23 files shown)


thostetler (changes requested):
Instead of using `<i>` tags with a listener, can we switch to a `<button>` with the icon inside

Something like this:
```
<button class="btn btn-link download">
  <i class="fa-download ..." aria-hidden="true"></i>
  <span class="sr-only">download data</span>
</button>
```
which would have better accessibility

shinyichen:
It's been updated as suggested.

> Instead of using `<i>` tags with a listener, can we switch to a `<button>` with the icon inside
> 
> Something like this:
> 
> ```
> <button class="btn btn-link download">
>   <i class="fa-download ..." aria-hidden="true"></i>
>   <span class="sr-only">download data</span>
> </button>
> ```
> 
> which would have better accessibility

shinyichen:
@thostetler Now it's ready for review

thostetler (approved):
This LGTM, there may be a better way to do the actual download part.  I'm not positive if this style of adding attributes to a button and clicking is fully cross-browser -- but for now, I think this will work fine.

PR #2178, opened 2021-06-10
Bug fix: Results graph hanging after click on "close"

shinyichen:
To reproduce the problem, do a search, then go to "results graph". Change to the third type display (read count + citation count). Then click the close button. This will cause the component to spin forever and not able to close. Because when selected the third display, the x, y values have changed from default, then a call to reset (when click on close)  causes x and y values change back to default, causing a re-render call with no valid plot data.

This PR fixes this issue but preventing a re-render when closing.

Screen Shot 2021-06-10 at 6 32 29 AM


thostetler (approved):
LGTM

PR #2179, opened 2021-06-11
Fixed dark mode setting logic error.

shinyichen:
The error fixed: Dark mode is always set to use system setting if app setting is dark mode off.


thostetler (approved):
LGTM, this should fix the issue

PR #2185, opened 2021-07-01
"All Search Terms" dropdown description

shinyichen:
Using Bootstrap popover to show description of each search term when a dropdown option is focused.
Not shown in mobile and tablet

Screen Shot 2021-07-01 at 12 40 10 PM
Screen Shot 2021-07-01 at 12 44 12 PM
Screen Shot 2021-07-01 at 12 45 33 PM


thostetler (approved):
Sorry about the delay.

This LGTM, I'll merge

PR #2199, opened 2021-08-27
Update Node version to latest LTS

thostetler:
Also add back in `grunt/requirejs.js` so that build passes.

Removes unnecessary `'use strict';` directives that were constantly giving warnings


shinyichen (approved):
Tested and no problem.

PR #2208, opened 2021-09-17
Fix myADS dropdown menus

thostetler:
Switch to using different lib for menubutton
Small fixes to address errors in myADS menus
Adds styles based on bootstrap dropdown menu


thostetler:
@shinyichen I'm not sure if you originally add this `overflow-y: hidden` during your updates for darkmode, but I found it was causing the dropdown menu to fall behind the footer when opened.
https://github.com/thostetler/bumblebee/commit/5e351790bc1a16f19f1ba1cd9dff88e881937d2f#diff-29f40ee360e6aa05a31806652cafdb5d6d5e14307807a69eb7fda7e1b4ba7da1L102

Beforehand I had "fixed" this by just switching to a `dropup` for the last few, but this actually fixes it now.  Let me know if you think it'll cause any issues, or if there is a better way to handle it.

thostetler:
Also, I did my best to emulate the bootstrap `dropdown-menu` I didn't change much.  But please take a look at my css changes, and let me know if you think there could be improvements or whatever.  Thanks!

shinyichen:
> @shinyichen I'm not sure if you originally add this `overflow-y: hidden` during your updates for darkmode, but I found it was causing the dropdown menu to fall behind the footer when opened.
> https://github.com/adsabs/bumblebee/pull/2208/files#diff-29f40ee360e6aa05a31806652cafdb5d6d5e14307807a69eb7fda7e1b4ba7da1L102
> 
> Beforehand I had "fixed" this by just switching to a `dropup` for the last few, but this actually fixes it now. Let me know if you think it'll cause any issues, or if there is a better way to handle it.

It looks like it was added 6 years ago, I can't tell what it was trying to do.

shinyichen (approved):

PR #2211, opened 2021-09-20
Updated search examples

shinyichen:
* Use random search examples
* Removed three links on landing page


src/js/react/Recommender/components/SearchExamples.jsx.js
```diff
@@ -54,51 +54,56 @@ define([
     tooltip: PropTypes.string,
   };
 
-  const SearchExamples = () => {
+  const SearchExamples = React.memo(() => {
```
thostetler, line 57:
```suggestion
  const SearchExamples = () => {
```
I think we can get rid of React.memo here since there aren't any props to compare it against, it'll just re-render everytime anyway.

thostetler (approved):

src/js/react/Recommender/components/SearchExamples.jsx.js
shinyichen: I added this because it was updating the examples when I click on one of the examples. But it feels very odd to see the examples change when clicked on one of them.

shinyichen: Let me know if there's better way to handle it.

PR #2228, opened 2022-01-21
Merge resources and related materials

shinyichen:
* Combine the two components' UI into one container
* Use 'toggle' to show and hide long list of related materials
* Limit the height of the container, and if content is too long use scroll bar

### Normal view
Screen Shot 2022-01-21 at 11 30 03 AM
### Long list of related materials
Screen Shot 2022-01-21 at 11 28 00 AM
### On small screen
Screen Shot 2022-01-21 at 5 00 02 PM


shinyichen:
@thostetler Ops I still need to fix the tests. Please hold off on reviewing.

PR #2229, opened 2022-01-23
Improve resource timeout issue

thostetler:
As a stopgap measure, this PR adds a seconds CDN fallback URL.  For the
time-being removes SRI for external links.
Also improves app loading messages


src/js/components/application.js
```diff
@@ -406,13 +424,11 @@ define([
       return lazyLoad ? run : run();
     },
 
-    _setTimeout: function(deferred) {
```
shinyichen, line 409: Looks like tests still needs this function.

shinyichen: Also I am seeing the missing function error in console log in dev

shinyichen (changes requested):
The app does not start for me, it's looking for the `_setTimeout` function. Also in tests.

thostetler:
@shinyichen thanks for checking it out -- should hopefully be all fixed now.  Added a few updates also to the optimizer

shinyichen (approved):
I tested and didn't have any issues.

PR #2231, opened 2022-01-27
Added 'inst' to search quick field. 

shinyichen:
* Added `inst` to search quick field.
* Use a link in its popover description to a full list of institutions, which required changes the logic in closing the popover to handle link clicks.

### DIFF
--- a/src/js/widgets/search_bar/search_bar_widget.js
+++ b/src/js/widgets/search_bar/search_bar_widget.js
@@ -134,14 +134,16 @@ define([
       };
 
       // on close, move focus to search bar.  If we change page layout, may have to change this
-      $select2Instance.on('close', () => {
+      // delayed to handle user click on a link in the description before closing
+      $select2Instance.on('close', async () => {
         document.getElementById('query-search-input').focus();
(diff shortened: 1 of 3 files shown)


src/js/widgets/search_bar/search_bar_widget.js
```diff
@@ -134,14 +134,16 @@ define([
       };
 
       // on close, move focus to search bar.  If we change page layout, may have to change this
-      $select2Instance.on('close', () => {
+      // delayed to handle user click on a link in the description before closing
+      $select2Instance.on('close', async () => {
         document.getElementById('query-search-input').focus();
+        await new Promise((resolve) => setTimeout(resolve, 10));
```
thostetler, line 140: any reason to use async/await here? versus just doing the `remove()` in it's own `setTimeout` ?

shinyichen: You are right. I was thinking of the `sleep()` method in Java.

shinyichen: Updated.

PR #2242, opened 2022-02-23
Fix the public library getting into my libraries collection bug

shinyichen:
`fetchLibraryMetadata(id)` merges that library into your library collection. Which is what happens when you open a public library, but not what we want to do. So my fix was to only merge if it's not from viewing a public library.


thostetler (approved):
makes sense, LGTM

PR #2245, opened 2022-03-25
Show number of records found in library

shinyichen:
Show a warning when library's record count doesn't match number of records found.
Screen Shot 2022-03-25 at 10 01 29 AM

### DIFF
--- a/src/js/widgets/library_list/widget.js
+++ b/src/js/widgets/library_list/widget.js
@@ -323,6 +323,15 @@ define([
         // otherwise show a message urging users to add to collection
         this.model.set('hasRecords', true);
       }
+
+      // number of records actually found
+      this.model.set('numRecords', resp.metadata.num_documents);
+      this.model.set('numFound', resp.solr.response.numFound);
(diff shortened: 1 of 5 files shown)


src/js/widgets/library_list/widget.js
```diff
@@ -323,6 +323,15 @@ define([
         // otherwise show a message urging users to add to collection
         this.model.set('hasRecords', true);
       }
+
+      // number of records actually found
+      this.model.set('numRecords', resp.metadata.num_documents);
```
thostetler, line 328: I believe you can do it as an object. 

```
this.model.set({
  Foo: 'bar',
  Buzz: 'fizz'
})
```

thostetler (approved):
Looks good, I like the alert.  Couple suggestions in the comments

src/js/widgets/library_list/widget.js
shinyichen: Thanks for the tip. Updated

PR #2253, opened 2022-08-30
Rewording facet graphs description

thostetler:
Changes the wording for the graph facets descriptions (headers)

Adds a couple items to the model, switches the template to use flex and removes moves the text fully to the template.

image
image
image
image


kelockhart (approved):
Wording looks good! Let's double check the math with some testing in dev, but otherwise we're good to go.

thostetler:
Sounds good.  I'll merge and then we can deploy to dev for testing

PR #2285, opened 2024-02-15
More GA4 Updates

thostetler:
- Update Gtag on 404 and 500 pages
- Cover additional cases in analytics
- Update analytics calls in Resources
- Fix timing calls
- Consolidate and fix resource quick link calls
- Fix the export component route handler
- Couple of changes to some analytics calls
- Fix bug in classic_form property list
- Attempt to pass a good user_id to GA

### DIFF
--- a/src/js/widgets/abstract/widget.js
+++ b/src/js/widgets/abstract/widget.js
@@ -272,14 +272,7 @@ define([
       return false;
     },
 
-    emitAnalytics: function(e) {
-      // analytics(
-      //   'send',
-      //   'event',
-      //   'interaction',
-      //   'abs-full-text-link-followed',
-      //   e.target.dataset.target
-      // );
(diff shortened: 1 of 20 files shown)


src/js/widgets/abstract/widget.js
```diff
@@ -273,13 +273,13 @@ define([
...
 
     emitAnalytics: function(e) {
-      analytics(
-        'send',
-        'event',
-        'interaction',
-        'abs-full-text-link-followed',
-        e.target.dataset.target
-      );
+      // analytics(
```
shinyichen, line 276: was this intentionally commented out?

thostetler: Yeah... actually gonna make one small commit to delete that and update another call.

shinyichen (approved):

PR #2305, opened 2024-11-12
SCIX-490 - Fix ORCiD Bulk Claim Issue

thostetler:
There were two issues with how this bulk claim/delete calls were being made.

1. The collection from which it was getting the models to run the claims/deletes on were only the currently shown ones (i.e. active views).
  - I changed this so that it now considers all known collections (i.e. hiddenCollection)

2. It makes it's calls by sending an event through the active view and causing it to trigger an action.  This allows the UI to properly update to represent the current pending status.  However, if an active view is not found it was skipped.
  - I changed this to call the internal request method on all skipped view updates so it won't miss ones that were selected.

In the future this could be refactored further if we wish to be less coupled to the UI components.

### DIFF
--- a/src/js/modules/orcid/extension.js
+++ b/src/js/modules/orcid/extension.js
@@ -52,8 +52,6 @@ define([
           }
         });
 
-        console.group('Orcid Action');
-        console.log(`Running action: ${action} on ${models.length} models`);
         // go through each model and grab the view for triggering
         _.forEach(
           models,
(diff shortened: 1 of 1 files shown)


src/js/modules/orcid/extension.js
```diff
@@ -58,15 +65,23 @@ define([
               var view = this.view.children.findByModel(m);
 
               if (view) {
+                console.log('Found view for model, firing event on view', m);
```
shinyichen, line 68: Did you mean to keep the console log?

thostetler: Nope, you're right need to remove

